PR #87, opened 2018-07-23
Specify more clearly what the variableDuration VAST attribute means.

ryanthompson591:
This clarifies what the variableDuration attribute means.  It is a way for the Ad to tell the player what its expectations are.  Players that do not allow the ad to control playback should error out and not try to play ads that require pausing and extension.

Here is more discussion.
See https://github.com/InteractiveAdvertisingBureau/SIVIC/issues/20.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -845,7 +845,7 @@ parameters:
 
 #### resolve #### {#sivic-creative-clickThru-resolve}
 Describes the state of the player after click:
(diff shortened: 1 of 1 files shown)


ryanthompson591:
Change should to must, and then merge in.

There are also two ands at the end to fix.

ryanthompson591:
Dropping this as it is correctly merged in PR/248

PR #114, opened 2018-10-17
Change language for SIVIC:Player:init section

andmig:
Current section does not necessarily reflect creative engagement flow. Also it contradicts the fact that `resolve` is expected (or this is an information only message and `resolve` may be useless). See added issues.


ryanthompson591:
I think this is outdated, so I'm closing it.

PR #125, opened 2018-10-17
Section "Messages from the video element" language changes.

andmig:


ryanthompson591:
I think this is outdated and will close for now.

PR #277, opened 2019-06-20
Made a wide variety of edits

graphicdeveloper:
These are mostly aimed at filling out details and trying to make things more understandable to a less-technical audience. Got up to section 4, with edits also to section 6. Still more to do.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 27, 2019, 6:40 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FInteractiveAdvertisingBureau%2FSIMID%2Feffd7bf02e7516023f02b6b736b5e11a5f3b299e%2Findex.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 1.
FATAL ERROR: Unknown inline tag 'resolve' found:
  resolve` response before unloading the creative iframe. The player should establish a reasonable timeout to allow for creative ad end logic completion.&lt;/p>

&lt;h5 id='sivic-player-adSkipped-resolve' line-number=774>resolve&lt;/h5>
Creative finalizes internal ad end logic and responds with `
 ✘  Did not generate, due to fatal errors
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20InteractiveAdvertisingBureau/SIMID%23277.)._
</details>


graphicdeveloper:
I should add, I defer to the wisdom of the group in all cases. These are simply suggestions I'm making with the intention of adding clarity and making the spec readable to a wider audience than purely technical folks. I will take no offense to any feedback or rejections offered. Thanks!

index.bs
```diff
@@ -168,46 +173,43 @@ Example:
 </MediaFiles>
 ```
 
-Because the creative is an HTML resource that originates in the advertiser’s
-domain, interactive component loading involves creating a cross-origin iframe.
-This means bi-directional communication between the player and creative
-requires using the postMessage API. This, in turn, means that all
-communication between player and creative is asynchronous.
-
```
graphicdeveloper: Editors note: The above paragraph felt out of place to me in this section, since it's part of the Intro section and this is fairly technical. It also felt a bit out of context. I moved it into the "Player and Creative Communication" section (below), and also made some changes within the block of copy itself.

index.bs
```diff
@@ -168,46 +173,43 @@ Example:
...
 
 ## Scope and Limitation ## {#scope}
 
 The use of HTML is only required for the SIMID creative, not the publisher
 property hosting that creative. As long as the publisher can load HTML and
-communicate with it over the standard postMessage interface, it can support
+communicate with it over the standard postMessage API, it can support
 SIMID. In practice, this means that SIMID can be hosted in web page iframes,
-mobile app web views, and other platforms. In fact, SIMID can better support mobile use cases since SIMID can be used with the native player.
+mobile app web views, and other platforms. In fact, SIMID can better support mobile use cases than VPAID because SIMID has the ability to communicate with the native player via postMessage.
```
graphicdeveloper: Editor's Note: I'm not actually certain me above edit is technically correct: `... SIMID has the ability to communicate with the native player via postMessage.`  I was trying to clarify the prior version to be more clear, and assuming SIMID actually _can_ communicate with a native device player I figured it _has_ to be via postMessage, right? Please correct me if I'm wrong, which might mean that entire sentence should be removed?

index.bs
```diff
@@ -1150,17 +1152,21 @@ to render as though it was not stopped.
...
+
+Media players that do support the SIMID API should handle version negotiation between the creative and the media player via the
+[[#api-ad-loading]] algorithm. The SIMID API version is not identified by any element or attribute in the VAST file.
+
+A third, optional attribute which may be included on the InteractiveCreativeFile element is `variableDuration="true"`.
+If present, this attribute indicates that the ad unit is only playable if
+the media player allows the creative to pause playback of the ad's audio or video and extend the duration of the ad break (for example, with
+interactive content such as a game or survey). If the player does not support or allow this capability, then it **must not** render
+the current ad's audio/video or SIMID creative. The player should error out the ad instead (and either resume its primary content or continue
+on to the next ad in the current ad pod).
```
graphicdeveloper: The above section felt a bit thin to me as it was, and it also seemed like the proper place to move some of the technical details that were in the Intro section (originally with the <InteractiveMediaFile> example code). I hope you'll agree that this all works, but feel free to suggest revisions.

index.bs
```diff
@@ -1150,17 +1152,21 @@ to render as though it was not stopped.
...
-`apiFramework` attribute in the `InteractiveCreativeFile` element must be
-`SIMID` (all caps). This attribute identifies the SIMID API for the creative.
-Version negotiation should be handled by the [[#api-ad-loading]] algorithm
-(rather than identified in the VAST file).
-
-Another attribute of the `InteractiveCreativeFile` is `variableDuration`.
-If `variableDuration` is true this means that the ad is only playable if
-the player allows the ad to pause and extend duration. If the player does
-not support this capability, then it must not render the video or creative.
-The player should error out instead.
```
graphicdeveloper: Editor's Note: Is the 2nd-to-last sentence above *really* something we all agree on and are mandating for SIMID compliance from a media player? `...If the player does
not support this capability, then it must not render the video or creative.` ? Assuming yes, I added bold to "**must not**" in my edit below. I'm calling it out because this feels to me like something we should maybe draw even more attention to, since it's a significant business rule/restriction, and I'm not sure how we enforce compliance or tracking of that. Thoughts?

index.bs
```diff
@@ -67,23 +67,23 @@ primary use case of interactivity.
...
-      <td valign="top" colspan="1" rowspan="1">Video and interactive overlay can be pre-cached.</td>
+      <td valign="top" colspan="1" rowspan="1">Audio or video asset and SIMID creative can both be pre-cached.</td>
     </tr>
     <tr>
-      <td valign="top" colspan="1" rowspan="1">Errors influence on performance</td>
-      <td valign="top" colspan="1" rowspan="1">Possible site-wide fatal errors can result in significant publisher site performance deterioration.</td>
-      <td valign="top" colspan="1" rowspan="1">Internal creative errors do not affect player DOM, directly. Risks are limited to interactive component performance only.</td>
+      <td valign="top" colspan="1" rowspan="1">Errors influence on performance or UX</td>
+      <td valign="top" colspan="1" rowspan="1">Fatal script errors from ad creative can result in significant player or publisher site performance and user experience deterioration.</td>
+      <td valign="top" colspan="1" rowspan="1">Fatal script errors from ad creative do not affect player or publisher site directly. Impact is limited to creative performance and user experience only.</td>
```
aronschatz:
```suggestion
      <td valign="top" colspan="1" rowspan="1">Fatal script errors from ad creative do not affect player or publisher site directly. Impact is limited to performance and user experience of the creative, only.</td>
```

index.bs
aronschatz: I'm also not sure if that's technically correct. This needs an engineer to confirm.

index.bs
aronschatz: I think this is good copy.

index.bs
gasperk: It is, but it's not about the communication per se. SIMID is designed to allow the video player to be in the host layer, which means native in apps, whereas VPAID put the responsibility to manage the video player into the creative. (Though even in that case it can be pushed into native layer, but it wasn't designed with that in mind.)

I wasn't sure how to write this in a more spec-proper language, so I didn't make a suggestion here. :)

index.bs
```diff
@@ -136,27 +136,32 @@ primary use case of interactivity.
...
 as a “SIMID creative” or “creative”.
 
-Compliance with SIMID requires support for all features discussed in this document, unless a feature is noted as optional. 
+Compliance with SIMID requires support for all features and behaviors specificied in this document, unless a given feature or behavior is explicitly designated as optional.
 Standard RFC language will be used. See <a href="https://tools.ietf.org/html/rfc2119">https://tools.ietf.org/html/rfc2119</a>
 for RFC 2119 for enforcement terminology used in this standard.
 
 ## SIMID Interactive Creative Nature ## {#sivic-interactive-creative-nature}
 
 <img width="100%" src="images/simid_diagram_3_assets.png" alt="A diagram showing SIMID where and how sivic video and html assets load.">
```
ryanthompson591: since you're fixing, can you change sivic to SIMID here?

amitshetty:
Thought it looked good overall. Agree with most of the comments up here - not sure how to get those suggested changes in. Would that be @graphicdeveloper making those changes and re-submitting a PR? Or a follow on PR?

index.bs
```diff
@@ -67,23 +67,23 @@ primary use case of interactivity.
...
-      <td valign="top" colspan="1" rowspan="1">Player manages media playback.</td>
+      <td valign="top" colspan="1" rowspan="1">Ad media asset management</td>
+      <td valign="top" colspan="1" rowspan="1">Creative manages ad media loading and playback (video).</td>
+      <td valign="top" colspan="1" rowspan="1">Player manages ad media loading and playback (audio or video).</td>
     </tr>
     <tr>
       <td valign="top" colspan="1" rowspan="1">Pre-caching</td>
       <td valign="top" colspan="1" rowspan="1">Only the VPAID script can be reliably pre-cached. Video asset cannot be pre-cached.</td>
-      <td valign="top" colspan="1" rowspan="1">Video and interactive overlay can be pre-cached.</td>
+      <td valign="top" colspan="1" rowspan="1">Audio or video asset and SIMID creative can both be pre-cached.</td>
```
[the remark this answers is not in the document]
graphicdeveloper: I know it's a couple more words, but I actually think it's better to call audio/video out explicitly here given that the sentence closely relates the media asset with the SIMID creative itself, and not every reader may be clear on the separation of those components those components in SIMID (I'm intentionally driving the point, in other words).

index.bs
```diff
@@ -92,23 +92,23 @@ primary use case of interactivity.
...
+      <td valign="top" colspan="1" rowspan="1">Players can pre-load media and creative assets due to maintaining full control over decisioning, loading and displaying of the ad unit. Ad decisioning latency is removed, risks of internal processes are minimized due to separate security sandboxes.</td>
     </tr>
     <tr>
       <td valign="top" colspan="1" rowspan="1">API</td>
-      <td valign="top" colspan="1" rowspan="1">Both player and creative must support specific javascript functions.</td>
-      <td valign="top" colspan="1" rowspan="1">Utilizes native to browsers postMessage API. No custom functions participate in player vs. creative relationships. Player and creative communicate via exchanging of messages based on special protocol.</td>
+      <td valign="top" colspan="1" rowspan="1">Both the player and creative must support specific javascript functions. Each component calls functions directly on the other in a shared security sandbox (insecure).</td>
```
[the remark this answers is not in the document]
graphicdeveloper: I'm torn on this change suggestion, Gasper - I realize I use "Javascript context" elsewhere in this table - but in this particular case I have 2 arguments against your suggestion: 1) On the next line (below) I refer to "separate security sandboxes" as a direct comparison to the "shared security sandbox" here. 2) By context, I actually mean the technical global scope in the other places I used it. Here I'm trying to refer to the overall window/iframe difference (as in the diagram above this table). If you feel strongly about it I'll make your suggested change, but those are my counterpoints.

index.bs
graphicdeveloper: I think you noticed after making this comment that I caught and changed that 'sivic' to 'simid' in a separate PR #275 (along with a bunch of other anchor links).
https://github.com/InteractiveAdvertisingBureau/SIMID/pull/275/files

index.bs
graphicdeveloper: OK, I rewrote this to focus on the control of the media asset (rather than the communication). Let me know if you think this works better:
`In fact, SIMID can better support mobile use cases than VPAID because a native app or media player directly controls loading and playback of a SIMID ad unit's
media asset (whereas a VPAID ad unit offers no direct access to or control of its internal media asset).`

index.bs
graphicdeveloper: Thanks, glad you approve! :-D

index.bs
```diff
@@ -67,23 +67,23 @@ primary use case of interactivity.
...
       <td valign="top" colspan="1" rowspan="1">Only the VPAID script can be reliably pre-cached. Video asset cannot be pre-cached.</td>
-      <td valign="top" colspan="1" rowspan="1">Video and interactive overlay can be pre-cached.</td>
+      <td valign="top" colspan="1" rowspan="1">Audio or video asset and SIMID creative can both be pre-cached.</td>
     </tr>
     <tr>
-      <td valign="top" colspan="1" rowspan="1">Errors influence on performance</td>
-      <td valign="top" colspan="1" rowspan="1">Possible site-wide fatal errors can result in significant publisher site performance deterioration.</td>
-      <td valign="top" colspan="1" rowspan="1">Internal creative errors do not affect player DOM, directly. Risks are limited to interactive component performance only.</td>
+      <td valign="top" colspan="1" rowspan="1">Errors influence on performance or UX</td>
+      <td valign="top" colspan="1" rowspan="1">Fatal script errors from ad creative can result in significant player or publisher site performance and user experience deterioration.</td>
```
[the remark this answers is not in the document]
graphicdeveloper: This is already one of the wordier lines, so I had trouble adding more detail without it feeling overloaded. I ended up adding a parenthetical to the end, using similar semantics as elsewhere, which hopefully covers it well enough: ` ... (due to shared security sandbox).`

graphicdeveloper:
I just pushed a new commit with 3 changes based on feedback from the group. I also merged one of Aron's suggested changes directly from this PR.

I pushed back on a couple comments and explained my thinking, but am happy to make those changes as well if folks overrule me on them. We can discuss on Thursday. ;-)

PR #335, opened 2020-04-06
Updated Executive Summary

katiestroudpro:
Made th update in the .bs file this time.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 6, 2020, 8:33 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FInteractiveAdvertisingBureau%2FSIMID%2F06e4b6bc8910b66088109c7e3bbb35a69202b2a2%2Findex.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 1.
FATAL ERROR: Line 805 isn't indented enough (needs 1 indent) to be valid Markdown:
"&lt;/div>"
 ✘  Did not generate, due to fatal errors
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20InteractiveAdvertisingBureau/SIMID%23335.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -384,24 +384,26 @@ Markup Shorthands: markdown yes, idl yes, dfn yes, markup yes
 
 # Executive Summary # {#exec-summary}
 
-Secure Interactive Media Interface Definition (SIMID) is a standard for providing rich interactivity in the context of
-streaming audio and video (media) ads. While the Video Ad Serving Template (VAST) standard addresses how publishers
-discover various metadata assets related to an ad campaign, SIMID addresses
-how the publisher’s media player should communicate and interface with a rich
-interactive layer and vice versa. As such, one can think of the SIMID creative
-as one of the assets listed in a VAST document.
+Secure Interactive Media Interface Definition (SIMID) is an API for handling 
+secure interactions between media players and streaming media (audio or video) ads. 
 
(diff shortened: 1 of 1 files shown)


amitshetty:
Thanks for the edits Katie!
imo, these are stylistic changes, and I cant really say I like this better than the older one. I'd prefer to not make such a chance without a reason to.
And there are things in the older version I kind of prefer (like the early mention of relationship to VAST4, SSAI etc). I would instead suggest a much smaller editorial change to fix things .

aronschatz:
I agree, this focuses too much on just the security aspect of the standard without mentioning the other reasons for its creation. I think there are some good lines that could be added to the old copy and not change the entire section.

PR #352, opened 2020-06-23
Refactors promises to be more simple to read

ryanthompson591:
The new workflow introduces 2 promises:
1. sessionCreatedPromise_ that resolves when the simid creative initializes a session.
2. resolveInitializationPromise_ that resolves when initialization is resolved.

The workflow should be:
1. Creative creates  a session.
2. Player calls SIMID:player:Init
3. Creative calls SIMID:Player:startCreative

However, these are all asynchronous so they may be called out of order. In order to keep these in order, this sample code uses promises.

Calling Init will 
1. wait for the session to be created (via sessionCreatedPromise_) 
2. then call SIMID:player:init

Calling playAd will:
1. Wait for the creative to respond to the initialization message (via resolveInitializationPromise_)
2. then call SIMID:player:startCreative

### DIFF
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -65,15 +65,35 @@ class SimidPlayer {
     /**
      * Resolution function for the session created message
      */
-    this.sessionCreatedResolve_ = null;
+    this.resolveSessionCreatedPromise_ = null;
 
     /**
      * A promise that resolves once the creative creates a session.
      * @private {?Promise}
      */
-    this.sessionCreated_ = new Promise((resolve, reject) => {
-      this.sessionCreatedResolve_ = resolve;
+    this.sessionCreatedPromise_ = new Promise((resolve, reject) => {
+      this.resolveSessionCreatedPromise_ = resolve;
     });
+
+    /**
+     * Resolution function for the ad being initialized.
+     */
+    this.resolveInitializationPromise_ = null;
+
+    /**
+     * Reject function for the ad being initialized.
+     */
+    this.rejectInitializationPromise_ = null;
+
+    /**
+     * A promise that resolves once the creative responds to initialization with resolve.
+     * @private {?Promise}
+     */
+    this.initializationPromise_ = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


examples/player/simid_player.js
```diff
@@ -65,16 +65,36 @@ class SimidPlayer {
     /**
      * Resolution function for the session created message
```
damargulis, line 66: Add a type here

examples/player/simid_player.js
```diff
@@ -65,16 +65,36 @@ class SimidPlayer {
...
     });
 
+    /**
+     * Resolution function for the ad being initialized.
+     */
+    this.resolveInitializationPromise_ = null;
+
+    /**
+     * Reject function for the ad being initialized.
+     */
```
damargulis, line 85: Types for these also

examples/player/simid_player.js
ryanthompson591: Done here and elsewhere.

examples/player/simid_player.js
ryanthompson591: Done.

examples/player/simid_player.js
```diff
@@ -132,7 +161,7 @@ class SimidPlayer {
    * @private
    */
   addListeners_() {
-    this.simidProtocol.addListener(ProtocolMessage.CREATE_SESSION, this.sendInitMessage.bind(this));
+    this.simidProtocol.addListener(ProtocolMessage.CREATE_SESSION, this.onSessionCreated_.bind(this));
     this.simidProtocol.addListener(CreativeMessage.REQUEST_FULL_SCREEN, this.onRequestFullScreen.bind(this));
     this.simidProtocol.addListener(CreativeMessage.REQUEST_PLAY, this.onRequestPlay.bind(this));
     this.simidProtocol.addListener(CreativeMessage.REQUEST_PAUSE, this.onRequestPause.bind(this));
```
Kaleab470, line 167: Maybe consider changing these bind statements to ES6 style too

PR #354, opened 2020-07-01
Updated base creative onInit function

ryanthompson591:
This allows users to override onInit and still have access to data changes.

### DIFF
--- a/examples/creatives/base_simid_creative.js
+++ b/examples/creatives/base_simid_creative.js
@@ -28,7 +28,7 @@ class BaseSimidCreative {
       muted: false,
       paused: false,
       volume: 0.5,
-      fullscreen: false //TODO add this to environment data in spec
+      fullscreen: false
     }
 
 
@@ -77,6 +77,7 @@ class BaseSimidCreative {
   /**
    * Receives init message from the player.
    * @param {!Object} eventData Data from the event.
+   * @protected
    */
   onInit(eventData) {
     this.updateInternalOnInit(eventData);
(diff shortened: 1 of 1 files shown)


examples/creatives/base_simid_creative.js
```diff
@@ -20,7 +20,17 @@ class BaseSimidCreative {
...
-    this.videoState = {};
+    this.videoState = {
+      currentSrc:'',
+      currentTime: -1, // Time not yet known
+      duration: -1, // duration unknown
+      ended: false,
+      muted: false,
+      paused: false,
+      volume: 0.5,
+      fullscreen: false //TODO add this to environment data in spec
```
damargulis, line 31: I think you can remove the todo now, looks like it is in the spec

examples/creatives/base_simid_creative.js
```diff
@@ -69,21 +79,22 @@ class BaseSimidCreative {
...
+    this.updateInternalOnInit(eventData);
+    this.simidProtocol.resolve(eventData, {});
+  }
+
+  /**
+   * Updates internal data on initialization call.
+   *
+   * Note: When overriding the onInit function and not wishing
+   * to always resolve, subclasses may instead use this function.
+   * @param {!Object} eventData Data from the event.
```
damargulis, line 91: `@protected` ?

examples/creatives/base_simid_creative.js
ryanthompson591: Done.

examples/creatives/base_simid_creative.js
ryanthompson591: Yeah, I labeled all these functions protected, since they are meant to be overridden.

ryanthompson591:
PTAL.

PR #356, opened 2020-07-08
Fix init reject bug

ryanthompson591:
Previously when a creative responded with reject to an init message this would cause an error because we were not listening for a reject message immediately.

Also:
- Fixes a bug where the parameters of the reject message were not being passed through.
- Logs the parameters from the reject message.

### DIFF
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -410,8 +410,7 @@ class SimidPlayer {
     this.hideAdPlayer_();
     this.adVideoElement_.src = '';
     this.destroySimidIframe();
-    // TODO don't comment next line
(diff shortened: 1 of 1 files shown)


examples/player/simid_player.js
```diff
@@ -255,10 +264,10 @@ class SimidPlayer {
     }
     const initPromise = this.simidProtocol.sendMessage(
         PlayerMessage.INIT, initMessage);
-    initPromise.then(()=> {
-      this.resolveInitializationPromise_();
-    }).catch(() => {
-      this.rejectInitializationPromise_();
+    initPromise.then((args)=> {
```
juliareichel, line 267: Looks good!

PR #362, opened 2020-07-23
New files for nonlinear ad.

juliareichel:
-Created html and js files for non-linear ad
-Set up the two files


examples/creatives/nonlinear_overlay.html
```diff
@@ -0,0 +1,13 @@
+<html>
+    <head>
+      <script src="../simid_protocol.js"> </script>
+      <script src="base_simid_creative.js"> </script>
+      <script src="nonlinear.js"> </script>
```
kristen-m, line 5: Don't forget to rename

PR #363, opened 2020-07-23
Support non linear

Kaleab470:


examples/player/simid_player.js
```diff
@@ -10,7 +10,7 @@ class SimidPlayer {
    * from the creative.
    * @param {!Function} This function gets called when the ad stops.
```
damargulis, line 11: add jsdoc

examples/player/simid_player.js
```diff
@@ -144,8 +157,12 @@ class SimidPlayer {
     // The target of the player to send messages to is the newly
     // created iframe.
     playerDiv.appendChild(simidIframe);
-    // Set up css to overlay the SIMID iframe over the video creative.
-    simidIframe.classList.add('simid_creative');
+    // Set up css to overlay the SIMID iframe over the entire video creative
+    // only if linear.
+    if (this.isLinearAd_){
+      simidIframe.classList.add('simid_creative');
```
damargulis, line 163: Maybe should rename this class then to linear_simid_creative

damargulis, line 163: Is there any css needed for a non linear creative?

examples/player/simid_player.js
```diff
@@ -430,11 +452,14 @@ class SimidPlayer {
   
   /** The creative wants to play video. */
   onRequestPlay(incomingMessage) {
```
damargulis, line 454: You should probably reject if non linear

examples/player/simid_player.js
```diff
@@ -430,11 +452,14 @@ class SimidPlayer {
...
-      // The play function returns a promise.
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play().then(
+        // The play function returns a promise.
+        this.simidProtocol.resolve(incomingMessage),
+        this.simidProtocol.reject(incomingMessage)
```
damargulis, line 460: I know you didn't add these, but shouldn't they be `() => this.simidProtocol.resolve(message)`?

examples/player/simid_player.js
Kaleab470: No, it will instead be positioned by the given dimension inputs

examples/player/simid_player.js
Kaleab470: Done!

examples/player/simid_player.js
Kaleab470: Yeah, I cleaned it up!

examples/creatives/nonlinear_overlay.html
```diff
@@ -0,0 +1,13 @@
+<html>
```
ryanthompson591, line 1: WDYT about calling this nonlinear_QA or nonlinear_testersAd as really everything that is a nonlinear is an overlay.

examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * A sample SIMID nonlinear ad that shows an overlay of
+ * buttons to showcase functionality
+ */
+class NonLinear extends BaseSimidCreative {
```
ryanthompson591, line 5: Let's call this something other than nonlinear, then name the class similarily to how you name the file.

examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * A sample SIMID nonlinear ad that shows an overlay of
```
ryanthompson591, line 2: This needs to be commented differently than you might comment production code. Because this is sample code we can be a little more verbose.


An ad built for testing relevant SIMID API for nonlinear SIMID creatives. These API include:
expandCreative - This should expand the creative to take up the entire frame.
collapseCreative - Returns the creative to its original size.
resizeCreative - Allows a specific size for the creative.

examples/player/simid_player.html
```diff
@@ -78,9 +79,18 @@
...
         <button onclick="switchCreative('selector.html')">selector</button>
         <button onclick="switchCreative('extender.html')">extender</button>
+        <button onclick="switchCreative('nonlinear_overlay.html')">nonlinear_overlay</button>
       </td>
       </tr>
-
+      <tr>
+        <td>Ad type</td>
+        <td>
+          <input type="radio" id="linear_ad" name="ad_type" checked> <label for="linear_ad" style="font-size: 16px;">Linear
```
ryanthompson591, line 88: Lets make this radio button a little more active. When non linear is selected hide video url.  When in linear mode, show url.

examples/player/simid_player.html
```diff
@@ -78,9 +79,18 @@
         <button onclick="switchCreative('survey.html')">survey</button>
         <button onclick="switchCreative('selector.html')">selector</button>
         <button onclick="switchCreative('extender.html')">extender</button>
+        <button onclick="switchCreative('nonlinear_overlay.html')">nonlinear_overlay</button>
```
ryanthompson591, line 82: Add to this file something that sets defaults of all the UI when we select this creative, including the linear/non linear radio button.

examples/player/simid_player.js
```diff
@@ -144,8 +158,12 @@ class SimidPlayer {
     // The target of the player to send messages to is the newly
     // created iframe.
     playerDiv.appendChild(simidIframe);
-    // Set up css to overlay the SIMID iframe over the video creative.
-    simidIframe.classList.add('simid_creative');
+    // Set up css to overlay the SIMID iframe over the entire video creative
```
ryanthompson591, line 161: Treat these comments as explanations to developers that might be reading this, move it right into the if statement.

Does this method even work of using css?  When you specify specifying a specific size how does this select the correct offsets etc.

examples/player/simid_player.js
```diff
@@ -271,9 +289,14 @@ class SimidPlayer {
     // Once the ad is successfully initialized it can start.
     // If the ad is not visible it must be made visible here.
     this.showSimidIFrame_();
-    this.showAdPlayer_();
-    this.adVideoElement_.src = document.getElementById('video_url').value;
-    this.adVideoElement_.play();
+
+    // only linear ones play ad videos
```
ryanthompson591, line 293: make this descriptive for developers.
Suggestion:
// A linear ad will pause the content video and show the ad video. A nonlinear ad will overlay the content video.

A possibly more readable way you could code this would be:

if (this.isLinearAd_) {
  playLinearVideoAd_();
}
....

/**
 * Pauses the content video and starts playback of the ad video.
 */
 playLinearVideoAd_() {
      this.contentVideoElement_.pause();
      this.showAdPlayer_();
      this.adVideoElement_.src = document.getElementById('video_url').value;
      this.adVideoElement_.play();
}

examples/player/simid_player.js
```diff
@@ -120,7 +127,14 @@ class SimidPlayer {
    * Plays a SIMID  creative once it has responded to the initialize ad message.
    */
   playAd() {
-    this.contentVideoElement_.pause();
+
+    // content video shouldn't be paused for non-linear ads
```
ryanthompson591, line 131: Why not move all this logic down below where we have the other
if (this.isLinear_) statement?

Keep the pausing and playing stuff together.

examples/player/simid_player.html
damargulis, line 88: try to avoid inline css, make this a class and add it to the stylesheet

examples/player/simid_player.js
```diff
@@ -9,8 +9,9 @@ class SimidPlayer {
    * Sets up the creative iframe and starts listening for messages
    * from the creative.
    * @param {!Function} This function gets called when the ad stops.
+   * @param {boolean} Represents if the ad is a linear one
```
damargulis, line 12: Format for these should be 
`@param {type} variableName Description.`

Make sure you add the variable name in and end with a period

examples/player/simid_player.js
```diff
@@ -55,6 +56,12 @@ class SimidPlayer {
      */
     this.adComplete_ = adComplete;
 
+    /**
+     * A boolean indicating whether type of creative
```
damargulis, line 60: missing some text here

examples/player/simid_player.js
```diff
@@ -55,6 +56,12 @@ class SimidPlayer {
...
+     * @private
```
damargulis, line 61: include `@const` if its not going to change, and include the type as well

examples/creatives/nonlinear_overlay.html
ryanthompson591: I was just looking at Julias PR and it seems like this will collide with what she is doing. Ideally you could just do the player changes in this PR and her PR could have all the creative changes.

examples/player/simid_player.js
```diff
@@ -430,11 +453,12 @@ class SimidPlayer {
...
-      // The play function returns a promise.
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play().then(() => this.simidProtocol.resolve(incomingMessage));
+    } else {
+      this.simidProtocol.reject(incomingMessage);
+    }
```
damargulis, line 461: You still want to reject in the case where `this.adVideoElement_.play()` rejects (that happens more often then you would think -- for example, if the videoElement is currently trying to pause, if the source fails to load, etc.)

examples/player/simid_player.html
Kaleab470: The input for the dimensions is added in the other PR so I'll add it there for the full functionality

examples/player/simid_player.html
Kaleab470: done

examples/player/simid_player.js
```diff
@@ -55,6 +56,13 @@ class SimidPlayer {
      */
     this.adComplete_ = adComplete;
 
+    /**
```
damargulis, line 59: minor nits. include a period after `is`
The tag is `@const` not constant
the const, private, and type can all go in the same line (`@const @private {boolean}`)

examples/player/simid_player.js
```diff
@@ -131,7 +139,7 @@ class SimidPlayer {
    * Plays a SIMID  creative once it has responded to the initialize ad message.
    */
   playAd() {
-    this.contentVideoElement_.pause();
```
damargulis, line 134: I like the thought of moving this to where the other video stuff happens, but by doing that I think this means that we will wait until the SIMID creative is ready before pausing the content.
I think this could cause problems if there is a slow connection.  For example, say there is supposed to be a commercial break at 0:30. At 0:30, playAd would be called, but say it takes 2 seconds for the simid ad to download and be ready. The init message would not resolve until 0:32, meaning 2 seconds of post-ad-break content would play, before it got paused and the ad started.

Although, since with this example code the ad is manually initialized before you manually click playAd, maybe this isnt a problem? 

@ryanthompson591 thoughts?

examples/player/simid_player.js
```diff
@@ -55,6 +56,13 @@ class SimidPlayer {
      */
     this.adComplete_ = adComplete;
```
kristen-m, line 57: Does this need a JSDoc?

examples/player/simid_player.js
```diff
@@ -440,11 +467,14 @@ class SimidPlayer {
...
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play()
+      .then(() => this.simidProtocol.resolve(incomingMessage))
+      .catch(() => this.simidProtocol.reject(incomingMessage));
+    } else {
+      this.simidProtocol.reject(incomingMessage);
```
ryanthompson591, line 476: You need to reject non linear requests.

if (!this.isLinearAd_) {
errorMessage = {
errorCode : 1102,
message: 'non linear ads do not play video'
}
this.simidProtocol.reject(incomingMessage, errorMessage);

examples/player/simid_player.js
```diff
@@ -440,11 +467,14 @@ class SimidPlayer {
...
-    this.adVideoElement_.play().then(
-      // The play function returns a promise.
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play()
+      .then(() => this.simidProtocol.resolve(incomingMessage))
+      .catch(() => this.simidProtocol.reject(incomingMessage));
```
ryanthompson591, line 474: now is probably a good time to add a reason for the reject message.

maybe use 1206 - The SIMID media could not be loaded.

examples/player/simid_player.js
Kaleab470: i think it can be confusing what adComplete_ is without the jsdoc. I could be wrong

examples/player/simid_player.html
```diff
@@ -90,7 +105,7 @@
       <tr>
         <td>Media File</td>
         <td><input type="text" id="video_url" size = 30 
-           value="../media/IAB480-small.mp4">
+           value="https://storage.googleapis.com/gvabox/media/red30s.mp4">
```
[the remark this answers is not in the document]
Kaleab470: done

examples/creatives/nonlinear_tester.html
```diff
@@ -2,7 +2,7 @@
     <head>
       <script src="../simid_protocol.js"> </script>
       <script src="base_simid_creative.js"> </script>
-      <script src="nonlinear_overlay.js"> </script>
+      <script src="nonlinear_tester.js"> </script>
```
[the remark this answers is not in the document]
Kaleab470: i removed the creative file and js

examples/player/simid_player.js
```diff
@@ -440,11 +468,24 @@ class SimidPlayer {
...
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play()
+      .then(() => this.simidProtocol.resolve(incomingMessage))
+      .catch(() => {
+        errorMessage = {
+          errorCode : 1206,
```
ryanthompson591, line 477: I suggest using 
PlayerErrorCode.VIDEO_COULD_NOT_LOAD

examples/player/simid_player.js
```diff
@@ -440,11 +468,24 @@ class SimidPlayer {
...
+          message: 'The SIMID media could not be loaded.'
+        }
+        this.simidProtocol.reject(incomingMessage, errorMessage);
+      });
+    } else {
+      errorMessage = {
+        errorCode : 1102,
```
ryanthompson591, line 484: I suggest using a constant here.

examples/player/simid_player.js
Kaleab470: done

examples/player/simid_player.js
Kaleab470: done

examples/player/simid_player.js
```diff
@@ -440,11 +468,24 @@ class SimidPlayer {
...
-      // The play function returns a promise.
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+    
+    if (this.isLinearAd_) {
+      this.adVideoElement_.play()
+      .then(() => this.simidProtocol.resolve(incomingMessage))
+      .catch(() => {
+        errorMessage = {
```
damargulis, line 476: `const errorMessage` and below
js is weird, itll work if you never actually declare the variable (by using let, var, or const), but itll put the variable into the global scope by default, which isn't ideal

Kaleab470: done

PR #364, opened 2020-07-23
Sizing iframe

juliareichel:
-If ad is non-linear dimensions should be specified
-If ad is linear dimensions will fill the entire player (as of now) and inputs for dimensions are disabled
-Allow users to resize the creative ad


examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
       <script src="nonlinear_overlay.js"> </script>
     </head>
     <style>
+      button {
```
damargulis, line 8: Did the other example ads do css this way or did they use an external style sheet?  If they did it this way then its fine to maintain this,  otherwise id say to use an external sheet.

examples/creatives/nonlinear_overlay.js
```diff
@@ -6,4 +6,20 @@ class NonLinear extends BaseSimidCreative {
     constructor() {
       super();
     }
+
+    onRequestResize() {
+      const resize_x_val = document.getElementById('resize_x_val').value;
```
damargulis, line 11: optional nit. if you name all of these the same thing as the key they will be used for below, you can define the object using shorthand properties

```
const x_val = element.value;
const params = {x_val};
```

examples/creatives/nonlinear_overlay.js
```diff
@@ -6,4 +6,20 @@ class NonLinear extends BaseSimidCreative {
...
+      const resize_y_val = document.getElementById('resize_y_val').value;
+      const resize_width = document.getElementById('resize_width').value;
+      const resize_height = document.getElementById('resize_height').value;
+  
+      const resize_params = {
+        'x_val': resize_x_val,
```
damargulis, line 17: https://interactiveadvertisingbureau.github.io/SIMID/#simid-creative-requestResize

The keys should be x and y, not x_val and y_val

examples/player/simid_player.html
```diff
@@ -23,15 +23,30 @@
     }
   </style>
   <script>
+    //if ad is non-linear allow user to insert specific dimensions
+    function enableDisable() {
```
damargulis, line 27: enable/disable what?

Also, imo use toggle instead of enableDisable

examples/player/simid_player.html
```diff
@@ -23,15 +23,30 @@
...
+      let isNonLin = document.getElementById("nonlinear_ad");
```
damargulis, line 28: In general, avoid abbreviations

https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

`isNonLinear`

But also, this variable is only the selector element, calling something prefixed with `is` makes it sound like its a boolean.  I would either make this variable the actual boolean, or name it something to indicate that it is an element

examples/player/simid_player.html
```diff
@@ -23,15 +23,30 @@
...
+      let uniqueDimensions = document.querySelectorAll(".unique_dimension");
+      uniqueDimensions.forEach(function(dimension) {
```
damargulis, line 30: `(dimensions) => {`

examples/player/simid_player.js
```diff
@@ -55,6 +55,12 @@ class SimidPlayer {
      */
     this.adComplete_ = adComplete;
 
+    /**
+     * A boolean indicating whether type of creative
```
damargulis, line 59: I think you're missing a couple words here

examples/player/simid_player.js
```diff
@@ -144,8 +162,12 @@ class SimidPlayer {
     // The target of the player to send messages to is the newly
     // created iframe.
     playerDiv.appendChild(simidIframe);
-    // Set up css to overlay the SIMID iframe over the video creative.
-    simidIframe.classList.add('simid_creative');
+    // Set up css to overlay the SIMID iframe over the entire video creative
+    // only if linear.
+    if (this.isLinearAd_){
+      simidIframe.classList.add('simid_creative');
```
damargulis, line 168: I think it makes sense to change the name of this css class, if its only for linear ads now to indicate as such

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
   }
 
   /**
-   * Returns the dimensions of an element within the player div.
+   * Returns the full dimensions of an element within the player div.
+   * These dimensions will equal the full size of the player.
```
damargulis, line 224: This description doesn't make sense to me.  Is it getting the dimensions of the element within the player div, or the dimensions of the entire player div?  Depending on the element, those could be different values

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+    const width = document.getElementById('width').value;
+    const height = document.getElementById('height').value;
+
+    return {
+      'x' : x_val,
+      'y' : y_val,
+      'width' : width,
+      'height' : height,
+      // TODO: This example does not currently support transition duration.
+      'transitionDuration': 0
```
damargulis, line 256: Maybe a question for ryan, but what is this? I don't see it anywhere in the spec

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
       // TODO: This example does not currently support transition duration.
       'transitionDuration': 0
     };
   }
 
+  /**
+   * Returns the specified dimensions of the non-linear creative.
+   * @return {!Object}
+   */
+  getNonLinearDimensions() {
```
damargulis, line 244: This function says its returing the dimensions of the non-linear creative, but its taking the values from the input buttons.  If you initialize the nonlinear creative, but then the creative requests a change in size, then you call this again, it will return the wrong values

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+    const dimensions = this.getNonLinearDimensions();
+    
+    this.simidIframe_.style.height = dimensions.height;
+    this.simidIframe_.style.width = dimensions.width;
+    this.simidIframe_.style.left = `${dimensions.x}px`;
+    this.simidIframe_.style.top = `${dimensions.y}px`;
+
+    this.simidIframe_.style.position = "absolute";
+
+    this.contentVideoElement_.play();
```
damargulis, line 274: This doesn't seem like it belongs in a function called "displayNonLinearCreative", as playing the video is a separate thing from "displaying" the creative

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+  }
+
+  /**
+   * Allows users to request resizing of the creative
+   * @param {!Object} incomingMessage Message sent from the 
+   *   creative to the player
+   */
+  onRequestResize(incomingMessage) {
+    const x_val  = incomingMessage.args['x_val'];
```
damargulis, line 283: Again, I think according to the spec these keys should be x and y

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+    const y_val  = incomingMessage.args['y_val'];
+    const width  = incomingMessage.args['width'];
+    const height  = incomingMessage.args['height'];
+
+    const resizeDimensions = {
+      'x' : x_val,
+      'y' : y_val,
+      'width' : width,
+      'height' : height,
+    }
```
damargulis, line 293: Why destruction the args object, just to recreate the same object?

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+
+    if (this.isLinearAd_) {
+      console.log("Cannot resize linear ad");
+      return;
```
damargulis, line 297: Should reject?

examples/player/simid_player.js
```diff
@@ -197,33 +220,100 @@ class SimidPlayer {
...
+    }
+
+    this.simidIframe_.style.height = resizeDimensions.height;
```
damargulis, line 300: Youre probably going to want to validate these values somewhere.  Make sure that they are actually numbers, and that they are valid numbers (not infinity, not negative, not bigger than the player etc.)

examples/player/simid_player.js
```diff
@@ -430,11 +525,14 @@ class SimidPlayer {
   
   /** The creative wants to play video. */
   onRequestPlay(incomingMessage) {
-    this.adVideoElement_.play().then(
-      // The play function returns a promise.
-      this.simidProtocol.resolve(incomingMessage),
-      this.simidProtocol.reject(incomingMessage)
-    );
+
```
damargulis, line 528: I think I might have already commented this on a different cl (maybe duplicating some commits?), but I think these should be `() => this.simidProtocol.resolve(message)`

examples/player/simid_player.html
```diff
@@ -23,15 +23,30 @@
     }
   </style>
   <script>
+    //if ad is non-linear allow user to insert specific dimensions
```
ryanthompson591, line 26: I'm a fan of doing all player changes in one PR and all creative changes in a different PR.  Theoreticaly these two elements should function independently of one another.

examples/player/simid_player.html
ryanthompson591, line 27: If I'm understanding this correctly, I don't think we should do this in the sample code. We should just allow the ad to change dimensions (if possible).

We are adding complexity we might not need.

examples/player/simid_player.html
```diff
@@ -23,15 +23,30 @@
...
+    }
+
     let simidPlayer;
     function initAd() {
       // clean old ad if it exists
       if (simidPlayer) {
         simidPlayer.stopAd();
       }
 
+      let isLinear = document.getElementById("linear_ad").checked;
```
ryanthompson591, line 47: I'm worried you added to this PR stuff that's in kaleabs pr.

Are you using
git commit -a -m "stuff"

Prefer using
git add file_I_want_to_add.js
git commit -m "a comment about this commit"


That way files you don't want to be part of the PR won't be there.

examples/creatives/nonlinear_overlay.html
juliareichel: the other ads do it this way- just tried to stay consistent

examples/player/simid_player.js
juliareichel: we do- Kaleab put that in a different PR

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+      </div>
+    </div>
     <script>
+      //The creative should first call ready
+       const nonLinear = new NonLinear();
+      // This implementation is immediately ready but other creatives
+      // may wish to wait for assets to load before calling ready.
+      nonLinear.ready();
+
+      function resizing() {
```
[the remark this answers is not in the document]
juliareichel: fixed

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
       <script src="nonlinear_overlay.js"> </script>
```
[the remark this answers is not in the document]
juliareichel: renamed

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+  
+        <label for="height">Height: </label>
+        <input type="number" id="resize_height" min="0" style="width: 50px;">
+  
+        <button id="resize_button" onclick="resizing()">Resize Ad</button>
+      </div>
+    </div>
     <script>
+      //The creative should first call ready
+       const nonLinear = new NonLinear();
```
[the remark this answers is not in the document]
juliareichel: fixed

examples/player/simid_player.html
juliareichel: when we originally spoke you mentioned only allowing specified sizing for non-linear ads. Would you rather allow sizing for linear as well?

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+        color: white;
+      }
     </style>
   <body>
+    <div id="content_container">
+      <div id="resizing_inputs">
+        <label for="x_val">X: </label>
+        <input type="number" id="resize_x_val" min="0" style="width: 50px;">
+  
+        <label for="y_val">Y: </label>
```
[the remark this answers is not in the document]
juliareichel: Fixed

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+        <input type="number" id="resize_y_val" min="0" style="width: 50px;">
+  
+        <label for="width">Width: </label>
+        <input type="number" id="resize_width" min="0" style="width: 50px;">
+  
+        <label for="height">Height: </label>
+        <input type="number" id="resize_height" min="0" style="width: 50px;">
+  
+        <button id="resize_button" onclick="resizing()">Resize Ad</button>
```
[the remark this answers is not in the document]
juliareichel: got rid of the function

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+      </div>
+    </div>
     <script>
+      //The creative should first call ready
```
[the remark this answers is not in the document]
juliareichel: changed

examples/creatives/nonlinear_overlay.js
juliareichel: fixed

examples/player/simid_player.html
juliareichel: changed this function in button PR

examples/player/simid_player.html
juliareichel: in the toggle function that replaces this in the other PR I don't even save the variable so this is fixed

examples/player/simid_player.html
juliareichel: fixed in toggle function

examples/player/simid_player.js
juliareichel: weird, must have gotten deleted. Fixed

examples/player/simid_player.js
juliareichel: Good spot- never changed it from original player

examples/player/simid_player.js
juliareichel: moved to startCreativePlayback function

examples/player/simid_player.js
juliareichel: I added the reject but the only way to prevent the actual iframe from resizing is by returning

examples/player/simid_player.js
juliareichel: Ryan said this was in an old version of the spec but is not actually being used. I deleted it

examples/player/simid_player.js
juliareichel: fixed

examples/player/simid_player.js
juliareichel: hm you're right. I don't know why I initially did it this way. Got rid of the destruction

examples/player/simid_player.js
juliareichel: this function returns the SPECIFIED dimensions meaning the original ones that were declared from initialization. Maybe I just need to make the comment clearer

examples/player/simid_player.html
juliareichel: it's because we are branching off of each others. Will try to fix and compare to his PR not master

examples/creatives/nonlinear_overlay.js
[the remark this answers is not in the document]
juliareichel: great idea

juliareichel: took Ryan's idea and just set the properties of the original empty object

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+      }
     </style>
   <body>
+    <div id="content_container">
+      <div id="resizing_inputs">
+        <label for="x_val">X: </label>
+        <input type="number" id="resize_x_val" min="0" style="width: 50px;">
+  
+        <label for="y_val">Y: </label>
+        <input type="number" id="resize_y_val" min="0" style="width: 50px;">
```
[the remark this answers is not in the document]
juliareichel: will add in the css PR

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
+        cursor: pointer;
+      }
+      label {
+        color: white;
+      }
     </style>
   <body>
+    <div id="content_container">
+      <div id="resizing_inputs">
+        <label for="x_val">X: </label>
```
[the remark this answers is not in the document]
juliareichel: will do

examples/creatives/nonlinear_overlay.html
```diff
@@ -5,9 +5,41 @@
...
     <style>
+      button {
+        cursor: pointer;
+      }
+      label {
+        color: white;
+      }
     </style>
   <body>
+    <div id="content_container">
```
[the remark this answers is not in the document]
juliareichel: will do in css PR

examples/player/simid_player.js
juliareichel: kaleab fixed

examples/player/simid_player.js
juliareichel: made comment more clear

examples/player/simid_player.html
```diff
@@ -23,6 +23,20 @@
     }
   </style>
   <script>
+    //if ad is non-linear allow user to insert specific dimensions
+    function enableDisable() {
```
ryanthompson591, line 27: Maybe call the function something like.

modifyControlVisibility

examples/player/simid_player.html
```diff
@@ -4,7 +4,7 @@
   <script src="../simid_protocol.js"></script>
   <script src="simid_player.js"></script>
   <style>
-    .simid_creative {
```
ryanthompson591, line 7: what do you think of just removing the simid_player changes from this PR entirely, since you might conflict with kaleabs changes.

examples/player/simid_player.html
juliareichel: this function is called setButtonState (it's changed in the buttons PR)

PR #366, opened 2020-07-23
Expand & collapse

Kaleab470:


examples/creatives/testers_nonlinear.js
```diff
@@ -28,9 +30,7 @@ class TestersNonLinear extends BaseSimidCreative {
         'click', sendMessageFunction.bind(this));
   }
 
-  /**
-   * When creative asks to resize itself send message to the player.
-   */
+  /** When creative asks to resize itself, this sends message to the player. */
```
ryanthompson591, line 33: this sends a message to the player

examples/player/simid_player.html
```diff
@@ -147,22 +153,38 @@
       <tr>
         <td>Ad type</td>
         <td>
-          <input type="radio" id="linear_ad" name="ad_type" checked> <label for="linear_ad" class="ad_type_label">Linear
+          <input type="radio" id="linear_ad" name="ad_type" onclick="toggleInputs()"> <label for="linear_ad" class="ad_type_label">Linear
```
ryanthompson591, line 156: before we were specifying that this was checked. Does this still start as checked by default?

examples/player/simid_player.html
```diff
@@ -147,22 +153,38 @@
...
+          <label for="height">Height: </label>
+          <input type="number" id="height" value="120" min="0" class="resize_inputs">
         </td>
       </tr>
-      <tr>
+      <tr id="media_file" class="hidden">
         <td>Media File</td>
         <td><input type="text" id="video_url" size = 30 
-           value="https://storage.googleapis.com/gvabox/media/red30s.mp4">>
+           value="https://storage.googleapis.com/gvabox/media/red30s.mp4">
```
ryanthompson591, line 181: I know I commented on this in the other PR but it didn't get changed. We don't want storage.googleapis.com stuff in the SIMID samples. Is it possible to just put this file right into your pr?

examples/player/simid_player.js
```diff
@@ -115,10 +115,19 @@ class SimidPlayer {
    * a session.
    */
   initializeAd() {
+
+    if (!this.isLinearAd_ && !this.isValidDimensions_(this.getCreativeDimensions_())) {
+      console.log('Dimensions bigger than player');
```
ryanthompson591, line 120: You could consider using alert here, but i think this is a case where it might be appropriate. To the casual user they won't know why their click was a no op.

Maybe be more verbose message would be good here too. "Unable to play a non linear ad with dimensions bigger than the player. Please modify dimensions to a smaller size."

examples/player/simid_player.js
```diff
@@ -240,67 +255,111 @@ class SimidPlayer {
...
+    this.simidIframe_.style.height = resizeDimensions.height;
+    this.simidIframe_.style.width = resizeDimensions.width;
+    this.simidIframe_.style.left = `${resizeDimensions.x}px`;
+    this.simidIframe_.style.top = `${resizeDimensions.y}px`;
+  }
+
+  /** The creative wants to expand the ad. */
+  onExpandResize(incomingMessage) {
+    if (this.isLinearAd_) {
+      this.simidProtocol.reject(incomingMessage, "Cannot expand linear ads");
```
ryanthompson591, line 300: Instead of a string use an object.
https://interactiveadvertisingbureau.github.io/SIMID/#protocol-message-categories-reject

examples/player/simid_player.js
```diff
@@ -115,10 +115,19 @@ class SimidPlayer {
    * a session.
    */
   initializeAd() {
+
+    if (!this.isLinearAd_ && !this.isValidDimensions_(this.getCreativeDimensions_())) {
+      console.log('Dimensions bigger than player');
+      return;
+    }
+
+    // After the iframe is created the player will wait until the ad
```
kristen-m, line 124: minor nit but should this comment be in `/** */ ` form?

examples/player/simid_player.js
```diff
@@ -240,67 +255,111 @@ class SimidPlayer {
...
    * on top of the video content.
    */
-  displayNonLinearCreative_() {
-    const dimensions = this.getNonLinearDimensions();
-    
-    this.simidIframe_.style.height = dimensions.height;
-    this.simidIframe_.style.width = dimensions.width;
-    this.simidIframe_.style.left = `${dimensions.x}px`;
-    this.simidIframe_.style.top = `${dimensions.y}px`;
+  displayNonLinearCreative(incomingMessage = CreativeMessage.REQUEST_COLLAPSE) {
```
ryanthompson591, line 316: Don't use this constant for a default incoming message.

An incoming message is an object with a code that lets it know which message it's responding to.

I think this will all be solved once you implement an onRequestCollapse function.

examples/player/simid_player.js
```diff
@@ -198,6 +207,8 @@ class SimidPlayer {
         this.onRequestChangeAdDuration.bind(this));
     this.simidProtocol.addListener(CreativeMessage.GET_MEDIA_STATE, this.onGetMediaState.bind(this));
     this.simidProtocol.addListener(CreativeMessage.LOG, this.onReceiveCreativeLog.bind(this));
+    this.simidProtocol.addListener(CreativeMessage.REQUEST_EXPAND, this.onExpandResize.bind(this));
+    this.simidProtocol.addListener(CreativeMessage.REQUEST_COLLAPSE, this.displayNonLinearCreative.bind(this));
```
ryanthompson591, line 211: took a little looking to figure this out. But i think you are ending up with some weird code (probably buggy) because you are trying to use the same function for setting up the non linear are you are for a request collapse call.  Better instead to just make a request collapse function.

examples/player/simid_player.js
```diff
@@ -240,67 +255,111 @@ class SimidPlayer {
...
   /**
    * Allows users to request resizing of the creative
-   * @param {!Object} incomingMessage Message sent from the 
-   *   creative to the player
+   * @param {!Object} incomingMessage Message sent from the creative to the player
    */
   onRequestResize(incomingMessage) {
+
     if (this.isLinearAd_) {
       this.simidProtocol.reject(incomingMessage, "Cannot resize linear ad");
```
ryanthompson591, line 341: can we change this error message to "linear resize not yet supported." Also change this from a string to an object.

https://interactiveadvertisingbureau.github.io/SIMID/#protocol-message-categories-reject

examples/player/simid_player.js
```diff
@@ -240,67 +255,111 @@ class SimidPlayer {
...
       console.log("Cannot resize linear ad");
-      return;
+    
+    } else if (!this.isValidDimensions_(incomingMessage.args.creativeDimensions)) { 
+      this.simidProtocol.reject(incomingMessage, "Dimensions bigger than player");
```
ryanthompson591, line 345: use object.

examples/player/simid_player.js
damargulis: Good thought, but we usually reserve `/** */` comments for jsdocs. Inline comments like this are fine to use the `//` syntax, or `/* */` if they're going to be very long (notice the jsdoc starts with 2 stars, a regular inline mult-line comment uses 1)

examples/player/simid_player.js
```diff
@@ -240,67 +255,111 @@ class SimidPlayer {
...
    * Initializes the SIMID creative with all data it needs.
    * @private
    */
   sendInitMessage_() {
-    const videoDimensions = this.getFullVideoDimensions(this.contentVideoElement_);
+    const videoDimensions = this.getFullDimensions_(this.contentVideoElement_);
     // Since the creative starts as hidden it will take on the
     // video element dimensions, so tell the ad about those dimensions.
-    const creativeDimensions = this.getFullVideoDimensions(this.contentVideoElement_);
+    const creativeDimensions = this.getFullDimensions_(this.contentVideoElement_);
```
ryanthompson591, line 362: Oh, I'm glad we saw this here. For non linears the creative dimensions will need to be the amount of space the creative has.

For example if our non linear is {x:5, y:5, width:20, height:20} we would need to pass that in here instead of the video dimensions.

examples/player/simid_player.html
juliareichel: ^ why get rid of where it defualts

examples/player/simid_player.js
juliareichel: I would just make the log more verbose, I don't think we should include alerts because another pop up seems more intrusive

examples/player/simid_player.html
Kaleab470: No, it doesn't start as checked. I could make linear checked on default though

examples/player/simid_player.js
Kaleab470: Done

examples/player/simid_player.js
```diff
@@ -284,6 +283,21 @@ class SimidPlayer {
     return newDimensions;
   }
 
+  /** Validates and displays the non-linear creative. */
+  displayNonlinearCreative_() {
+    const newDimensions = this.getCreativeDimensions_();
+    
+    if (!this.isValidDimensions_(newDimensions)) {
+      console.log('Unable to play a non-linear ad with dimensions bigger than the player. Please modify dimensions to a smaller size.');
```
juliareichel, line 296: there is a sendLog function in this file, perhaps utilize it here

ryanthompson591: I think sendLog might not work here. Because sendLog sends the message to the creative, but this might not be relevant to the creative.

If I'm not mistaken this message is related to the setup of the player only, and it's possible that the creative doesn't even exist yet.

examples/player/simid_player.html
```diff
@@ -132,22 +138,38 @@
       <tr>
         <td>Ad type</td>
         <td>
-          <input type="radio" id="linear_ad" name="ad_type" checked> <label for="linear_ad" class="ad_type_label">Linear
+          <input type="radio" id="linear_ad" name="ad_type" onclick="toggleInputs()" checked> <label for="linear_ad" class="ad_type_label">Linear
```
damargulis, line 141: If one of these buttons is already checked, but you click it again anyway, does this trigger the toggleInputs function?

IMO it would be safer to structure this more similar to how the setCreativeControlState function works, so it takes in the value for which it wants to set the button (https://github.com/InteractiveAdvertisingBureau/SIMID/blob/master/examples/player/simid_player.js#L419).

Toggle is good when the meaning of the actual button is to toggle something (something like a play/pause button, where the actual meaning of the button is "change the state from what it currently is").  In this case, you want a specific outcome when clicking, regardless of what the current state is, so probably best to encode that specifically.

examples/player/simid_player.js
Kaleab470: nothing is initialized at this point so can't use that

examples/player/simid_player.html
```diff
@@ -132,22 +138,38 @@
...
+        <td>Creative Dimensions</td>
+        <td>
+          <label for="x_val">X: </label>
+          <input type="number" id="x_val" value="5" min="0" class="resize_inputs">
+
+          <label for="y_val">Y: </label>
+          <input type="number" id="y_val" value="5" min="0" class="resize_inputs">
+
+          <label for="width">Width: </label>
+          <input type="number" id="width" value="500" min="0" class="resize_inputs">
```
damargulis, line 157: Just curious -- do these default values signify anything or are they just random?

juliareichel: we put them there so we didn't have to constantly input them manually

Kaleab470: these numbers are not too small/not too big and just hold the buttons and the input fields

examples/player/simid_player.js
```diff
@@ -233,7 +248,7 @@ class SimidPlayer {
    * Returns the full dimensions of an element within the player div.
    * @return {!Object}
    */
-  getFullVideoDimensions(elem) {
+  getFullDimensions_(elem) {
```
damargulis, line 251: add a private tag

examples/player/simid_player.js
```diff
@@ -245,67 +260,145 @@ class SimidPlayer {
   }
 
   /**
-   * Returns the original specified dimensions of the non-linear creative.
-   * @return {!Object}
+   * Checks whether the input dimensions are valid and fit in the player window.
+   * @param {!Object} dimensions A dimension that contains x, y, width & height fields.
+   * @return {boolean}
```
damargulis, line 265: private tag

examples/player/simid_player.js
```diff
@@ -245,67 +260,145 @@ class SimidPlayer {
   }
 
   /**
-   * Returns the original specified dimensions of the non-linear creative.
-   * @return {!Object}
+   * Checks whether the input dimensions are valid and fit in the player window.
+   * @param {!Object} dimensions A dimension that contains x, y, width & height fields.
```
damargulis, line 264: Looks like we pass an object that specifically has {x, y, width, height} through a few of these functions .. it might be good to formalize this as a class or a record type at least (can push this to a different pr, don't need to do as part of this one)

examples/player/simid_player.js
```diff
@@ -245,67 +260,145 @@ class SimidPlayer {
...
   /**
-   * Allows users to request resizing of the creative
-   * @param {!Object} incomingMessage Message sent from the 
-   *   creative to the player
+   * Changes the simid iframe dimensions to the given dimensions
+   * @param {!Object} A dimension that contains an x,y,width & height fields.
    */
-  onRequestResize(incomingMessage) {
+  setSimidIframeDimensions_(resizeDimensions) {
+    this.simidIframe_.style.height = resizeDimensions.height;
```
damargulis, line 311: why doesn't height and width need `px`?

examples/player/simid_player.js
```diff
@@ -245,67 +260,145 @@ class SimidPlayer {
...
+    this.simidIframe_.style.width = resizeDimensions.width;
+    this.simidIframe_.style.left = `${resizeDimensions.x}px`;
+    this.simidIframe_.style.top = `${resizeDimensions.y}px`;
+  }
+
+  /** The creative wants to expand the ad. */
```
damargulis, line 317: `@param`

examples/player/simid_player.js
```diff
@@ -245,67 +260,145 @@ class SimidPlayer {
...
+      this.contentVideoElement_.pause();
+      this.simidProtocol.resolve(incomingMessage);
     }
+  }
 
-    this.simidIframe_.style.height = incomingMessage.args.creativeDimensions['height'];
-    this.simidIframe_.style.width = incomingMessage.args.creativeDimensions['width'];
-    this.simidIframe_.style.left = incomingMessage.args.creativeDimensions['x'];
-    this.simidIframe_.style.top = incomingMessage.args.creativeDimensions['y'];
+  /** The creative wants to collapse the ad. */
```
damargulis, line 336: `@param`

examples/player/simid_player.js
Kaleab470: @juliareichel

PR #367, opened 2020-07-23
Button functionality

juliareichel:
-Added close button to the creative
-Added close ad button to the player
-Added fatal error button to the player
-Added skip button to the player
-Add disable/enable toggle-state of buttons


examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+class NonLinear extends BaseSimidCreative {
+    constructor() {
+      super();
+
+      this.addButtonClickActions_();
+    }
+
+    /**
+     * Adds actions to different buttons available on the overlay.
```
damargulis, line 13: `@private`

examples/creatives/nonlinear_overlay.html
```diff
@@ -0,0 +1,54 @@
...
+    <style>
+      button {
+        cursor: pointer;
+      }
+
+      #close_ad {
+      color: red;
+      border: solid white 1px;
+      font-size: 14;
+      float: right;
```
damargulis, line 16: indentation

examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    onRequestResize() {
+      const resize_x_val = document.getElementById('resize_x_val').value;
+      const resize_y_val = document.getElementById('resize_y_val').value;
+      const resize_width = document.getElementById('resize_width').value;
+      const resize_height = document.getElementById('resize_height').value;
+  
+      const resize_params = {
+        'x_val': resize_x_val,
```
damargulis, line 26: spec says x and y as keys

examples/player/simid_player.html
```diff
@@ -42,6 +42,24 @@
       simidPlayer.playAd();
     }
 
+    function closeAd() {
```
damargulis, line 45: Seems odd to immediately close the ad after init. Maybe these buttons just shouldn't be active until after the ad has been initialized -- then just to be safe you can early return here if simidPlayer doesnt exist

examples/player/simid_player.html
```diff
@@ -42,6 +42,24 @@
...
+
+    function skipAd() {
+      if (!simidPlayer) {
+        initAd();
+      }
+      simidPlayer.skipAd();
+    }
+
+    function fatalError() {
+      simidPlayer.fatalError();
```
damargulis, line 60: Why doesn't this one have a guard? What if simidPlayer doesn't exist when this is clicked?

examples/player/simid_player.html
```diff
@@ -100,10 +118,13 @@
...
     <button onclick="playAd()"> Start ad playback </button>
     <button onclick="initAd()"> Initialize the creative </button>
+    <button onclick="closeAd()"> Close the ad </button> 
+    <button onclick="skipAd()"> Skip ad </button>
+    <button onclick="fatalError()"> Fatal Error </button>
   </div>
 
   <div id="player_div" style="position:absolute;">
-    <video id="video_player" class="content" src="../media/Big_Buck_Bunny.mp4" controls> </video>
+    <video id="video_player" class="content" src="../media/fakeContent.mp4" controls> </video>
```
damargulis, line 127: Why switch? Is it just a smaller file?
Maybe we should get rid of the old media file then while were here?

examples/player/simid_player.js
```diff
@@ -131,6 +131,30 @@ class SimidPlayer {
     });
   }
 
+  /**
+   * Closes the ad and SIMID creative.
+   */
+  closeAd() {
+    this.onRequestStop(CreativeMessage.REQUEST_STOP);
+    this.sendLog("Ad closed by player");
```
damargulis, line 139: Am I missing something? I don't see a sendLog or onRequestStop method anywhere (or onRequestSkip or onCreativeFaltalError)

examples/player/simid_player.html
juliareichel: smaller file - the bunny one was making the player act weirdly

examples/player/simid_player.js
juliareichel: you don't? They are on line 477, 484, 528

examples/player/simid_player.html
juliareichel: fixed

examples/creatives/nonlinear_overlay.js
juliareichel: fixed

examples/creatives/nonlinear_overlay.js
juliareichel: added

examples/player/simid_player.js
damargulis: Ah yep I was missing something haha, I didn't realize they already existed so I didn't open up the non-changed lines

examples/creatives/nonlinear_overlay.html
juliareichel: fixed

examples/player/simid_player.js
damargulis: I don't think those are quite the methods you want to call tho.  If you look at those, they each take in an incomingMessage, which they then send to simidProtocol.resolve/reject.  Those are currently called when a message is received, and the message object is the argument.  You are passing in just the message name.  If you run this line, what message eventually gets sent to the creative? Is that the message you want to be sent?

examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Adds actions to different buttons available on the overlay.
+     * @private
+     */
+    addButtonClickActions_() {
+      this.sendMessageOnButtonClick_("close_ad", CreativeMessage.REQUEST_STOP);
+    }
+
+    onRequestResize() {
```
ryanthompson591, line 20: A lot of this code is mixed with code I've looked at in another PR.  Ideally this PR should only contain new changes.

examples/player/simid_player.html
```diff
@@ -42,6 +42,27 @@
...
+    function closeAd() {
+      if (!simidPlayer) {
+        initAd();
+      }
+      simidPlayer.closeAd();
+    }
+
+    function skipAd() {
+      if (!simidPlayer) {
+        initAd();
```
ryanthompson591, line 54: I don't get it.  So if the user clicks skip ad, he could end up initializing an ad and then immediately skipping it?

Seems like what you want is to just have the skip ad button not even available unless and ad has been initialized.  Then you could probably just get rid of this if statement.

examples/player/simid_player.html
```diff
@@ -42,6 +42,27 @@
...
+      }
+      simidPlayer.skipAd();
+    }
+
+    function fatalError() {
+      if (!simidPlayer) {
+        initAd();
```
ryanthompson591, line 61: Nope. We shouldn't init an ad here. same comment as above.

perhaps you could code it this way too.
if (simidPlayer) {
  simidPlayer.fatalError();
}

no reason to init an ad in this function.

examples/player/simid_player.html
```diff
@@ -100,10 +121,13 @@
     </table>
     <button onclick="playAd()"> Start ad playback </button>
     <button onclick="initAd()"> Initialize the creative </button>
+    <button onclick="closeAd()"> Close the ad </button>
```
ryanthompson591, line 124: what do you think about only having these show up if an ad is started.  Then could be greyed out or not visible.

examples/player/simid_player.html
ryanthompson591: Yeah, there was an issue with local python players not being able to handle the large video file.

examples/player/simid_player.html
juliareichel: fixed

examples/player/simid_player.html
juliareichel: okay that's what I originally did but based on PR comments I changed it. Will revert back

examples/player/simid_player.html
juliareichel: good idea, fixed this

examples/player/simid_player.html
juliareichel: fixed

examples/creatives/nonlinear_overlay.js
juliareichel: The reason for this is because I branched off of Kaleab's work but he didn't yet merge it into master

examples/player/simid_player.js
juliareichel: fixed to call proper functions

examples/player/simid_player.html
```diff
@@ -35,11 +35,39 @@
       simidPlayer.initializeAd();
     }
 
+    function toggleButtons() {
+      let adRequests = document.querySelectorAll(".ad_request");
+      adRequests.forEach(function(request) {
+        if (request.disabled == false) {
```
damargulis, line 41: You never need to compare things directly to boolean

`if (x == true)` can always be `if (x)`
`if (x == false)` -> `if (!x)`

or even better 
request.disabled = !request.disabled

or even better than that imo, would be to change this function to take in a param for if the buttons should be turned on or off, and set it to that value passed in. I think everytime you call this function, you have a specific desired result ..
Meaning for example, you call this function in skipAd. After skipAd is called, you always want to disable the buttons. Right now, this only works because you know that the skipAd button only shows when the button is active, but it would be a bit less brittle if in skipAd you called this function to specifically disable the buttons.  That way, you don't have to worry about the current state before calling this

examples/player/simid_player.html
```diff
@@ -35,11 +35,39 @@
...
     function playAd() {
       if (!simidPlayer) {
         initAd();
       }
       simidPlayer.playAd();
+      toggleButtons();
+    }
+
+    function closeAd() {
+      simidPlayer.playerCloseAd();
```
damargulis, line 59: imo these should still have a guard `if (!simidPlayer) {return}`
I know these buttons should only show up after simidPlayer is defined, but its good to code defensively whenever possible

https://en.wikipedia.org/wiki/Defensive_programming#:~:text=Defensive%20programming%20is%20a%20form,of%20software%20under%20unforeseen%20circumstances.&text=Defensive%20programming%20is%20an%20approach,of%20software%20bugs%20and%20problems.

examples/player/simid_player.js
```diff
@@ -131,6 +131,30 @@ class SimidPlayer {
     });
   }
 
+  /**
+   * Player closes the ad and SIMID creative.
+   */
+  playerCloseAd() {
+    this.stopAd();
+    this.sendLog("Ad stopped by player");
```
damargulis, line 139: I think the logs in these functions are redundant.  The stop ad message contains a stop code saying that it is a player initiated stop. Does sending the log provide any additional value?

examples/player/simid_player.js
```diff
@@ -393,6 +417,23 @@ class SimidPlayer {
...
+   * Skips the ad and destroys the ad iframe.
+   * @param {StopCode} reason The reason the ad will stop.
+   */
+  skipAd(reason = StopCode.PLAYER_INITATED) {
+    // The iframe is only hidden on ad stoppage. The ad might still request
+    // tracking pixels before it is cleaned up.
+    this.hideSimidIFrame_();
+    const closeMessage = {
+      'code': reason,
+    }
```
damargulis, line 430: This variable goes unused (I don't think you need it, skipAd doesn't take in a code the way a stopAd message does)
https://interactiveadvertisingbureau.github.io/SIMID/#simid-player-adSkipped

juliareichel: yeah I was a little confused that I didn't see SkipCodes in simid protocol. Got rid of this variable

examples/player/simid_player.html
juliareichel: okay, so I will add the guard as well

examples/player/simid_player.js
juliareichel: good for sample code to show the user the message sent from player

examples/player/simid_player.html
juliareichel: fixed toggle function to not depend on previous state

examples/player/simid_player.js
```diff
@@ -131,6 +131,31 @@ class SimidPlayer {
...
+  playerCloseAd() {
+    this.stopAd();
+    this.sendLog("Ad stopped by player");
+  }
+
+  /**
+   * Player skips the ad.
+   */
+  playerSkipAd() {
+    debugger;
```
damargulis, line 146: delete

examples/player/simid_player.js
```diff
@@ -393,6 +418,20 @@ class SimidPlayer {
         .then(() => this.destroyIframeAndResumeContent_());
   }
 
+  /**
+   * Skips the ad and destroys the ad iframe.
+   * @param {StopCode} reason The reason the ad will stop.
```
damargulis, line 423: delete

examples/player/simid_player.js
```diff
@@ -393,6 +418,20 @@ class SimidPlayer {
...
+   */
+  skipAd() {
+    // The iframe is only hidden on ad stoppage. The ad might still request
+    // tracking pixels before it is cleaned up.
+    this.hideSimidIFrame_();
+    // Wait for the SIMID creative to acknowledge stop and then clean
+    // up the iframe.
+    this.simidProtocol.sendMessage(PlayerMessage.AD_SKIPPED, "Ad skipped by player")
```
damargulis, line 431: The ad skipped message doesn't have any additional data. You don't need to pass in anything to sendMessage besides the message type

examples/player/simid_player.html
```diff
@@ -35,11 +35,43 @@
       simidPlayer.initializeAd();
     }
 
+    function disableButtons(state) {
```
damargulis, line 38: I would rework the naming here a bit.
Since this can now either disable or enable the buttons, something like setButtonState might convey that better.  Then instead of state as the param value, you can use disable, since that will specifically convey what aspect of the state this is setting.

examples/player/simid_player.html
```diff
@@ -35,11 +35,43 @@
...
+        request.disabled = state;
+      });
+    }
+
     function playAd() {
       if (!simidPlayer) {
         initAd();
       }
       simidPlayer.playAd();
+      disableButtons(false);
```
damargulis, line 50: When you use a hard-coded value as a param in a function call, we usually try to add an inline comment saying what that value is representing, so in this case

`disableButtons(/* state= */ false);`

examples/player/simid_player.js
juliareichel: done

examples/player/simid_player.js
juliareichel: done

examples/player/simid_player.js
juliareichel: when tracing through the calls this message would be added to the args field, but sure I can get rid of it if it is unnecessary

damargulis: Yeah, you're right it would be added to the message, but since the spec doesn't talk about this, any creative implementation wouldn't know to look for it

examples/player/simid_player.html
juliareichel: sounds good!

examples/player/simid_player.html
juliareichel: Cool! I've never seen this before

examples/creatives/nonlinear_overlay.html
```diff
@@ -0,0 +1,54 @@
+<html>
```
ryanthompson591, line 1: I thought we removed and renamed nonlinear_overlay to something else.

examples/creatives/nonlinear_overlay.js
```diff
@@ -0,0 +1,35 @@
...
+      const resize_x_val = document.getElementById('resize_x_val').value;
+      const resize_y_val = document.getElementById('resize_y_val').value;
+      const resize_width = document.getElementById('resize_width').value;
+      const resize_height = document.getElementById('resize_height').value;
+  
+      const resize_params = {
```
ryanthompson591, line 26: I think you made a bunch of changes to this in a separate PR.  Please use those updated changes.

examples/player/simid_player.html
```diff
@@ -35,11 +35,43 @@
       simidPlayer.initializeAd();
     }
 
+    function setButtonState(disable) {
```
ryanthompson591, line 38: can you document this function,  I'm a little confused by it.  Or better yet rename it.

examples/player/simid_player.html
```diff
@@ -35,11 +35,43 @@
...
       }
       simidPlayer.playAd();
+      setButtonState(/* state= */ false);
+    }
+
+    function closeAd() {
+      if (!simidPlayer) {
+        return;
+      }
+      simidPlayer.playerCloseAd();
```
ryanthompson591, line 57: I prefer just closeAd for a name.

The rule of thumb is this. If a class or object has a name -> Then you don't need to restate its name in the API.

For example:
class Foo{
  changeFooProperty()
}

Should instead be
class Foo{
  changeProperty()
}

Then you write
foo.changeProperty()

examples/player/simid_player.js
```diff
@@ -131,6 +131,30 @@ class SimidPlayer {
     });
   }
 
+  /**
+   * Player closes the ad and SIMID creative.
+   */
+  playerCloseAd() {
```
ryanthompson591, line 137: change to closeAd. When someone calls this it will be
player.closeAd()

no need to have it be
player.playerCloseAd()

Same with the other functions.

Or better yet, just get rid of this function entirely and call stopAd directly from wherever you are calling this.

examples/player/simid_player.js
```diff
@@ -131,6 +131,30 @@ class SimidPlayer {
...
+    this.stopAd();
+    this.sendLog("Ad stopped by player");
+  }
+
+  /**
+   * Player skips the ad.
+   */
+  playerSkipAd() {
```
ryanthompson591, line 145: If we can get rid of these functions that would be more readable I think. Did you have them here for debugging purposes?

examples/player/simid_player.js
```diff
@@ -393,6 +417,19 @@ class SimidPlayer {
...
   }
 
+  /**
+   * Skips the ad and destroys the ad iframe.
+   */
+  skipAd() {
+    // The iframe is only hidden on ad stoppage. The ad might still request
+    // tracking pixels before it is cleaned up.
+    this.hideSimidIFrame_();
+    // Wait for the SIMID creative to acknowledge stop and then clean
```
ryanthompson591, line 427: s/stop/skip

examples/creatives/nonlinear_overlay.html
juliareichel: yes we did in different PR, so I just pulled the changes into this

examples/player/simid_player.js
juliareichel: fixed

examples/player/simid_player.html
juliareichel: thank you for teaching me this

examples/player/simid_player.js
juliareichel: got rid of them

examples/player/simid_player.js
juliareichel: got rid of them

examples/player/simid_player.html
juliareichel: yes added comment

examples/player/simid_player.html
```diff
@@ -50,11 +50,49 @@
       simidPlayer.initializeAd();
     }
 
+    /* Disables or enables stop, skip, and fatal error buttons
```
damargulis, line 53: make this a proper js doc
```
/**
 * Description of the method.
 * @param {paramType} paramName Description of the param.
 */
```

examples/player/simid_player.html
```diff
@@ -50,11 +50,49 @@
...
+        request.disabled = disable;
+      });
+    }
+
     function playAd() {
       if (!simidPlayer) {
         initAd();
       }
       simidPlayer.playAd();
+      setButtonState(/* state= */ false);
```
damargulis, line 68: you changed the variable name to disable, so this should now be `/* disable= */`, same below

examples/player/simid_player.html
juliareichel: wasn't sure if I should do that in html file, but will change now

examples/player/simid_player.html
juliareichel: fixed that

examples/player/simid_player.html
```diff
@@ -50,11 +50,50 @@
...
       simidPlayer.playAd();
+      setButtonState(/* disable= */ false);
+    }
+
+    function closeAd() {
+      if (!simidPlayer) {
+        return;
+      }
+      simidPlayer.stopAd();
+      simidPlayer.sendLog("Ad stopped by player");
```
ryanthompson591, line 77: what does send log do?  It might not be too relevant to send a log to the creative if it already knows its been stopped.

examples/player/simid_player.html
ryanthompson591: How about,
setCreativeControlsState

I know it's a little more verbose but it might make more sense.

Then declare two constants at the top of this file
/** DISABLED and ENABLED constants are used to control the state of buttons. */
const DISABLED = true;
const ENABLED = false;

then when you call this function it will be
setCreativeControlsState(ENABLED);

super understandable.

examples/player/simid_player.html
juliareichel: This log is meant to distinguish between what action was actually sent by the player since Stop, Fatal Error, and Skip all shut down the ad. Let me know if you think I should get rid of the logs, I put them there to be informative

examples/player/simid_player.html
juliareichel: i will change the code to use consts

examples/creatives/testers_nonlinear.js
```diff
@@ -3,18 +3,42 @@
...
+   */
+  onRequestResize() {
+    const resize_params = {};
+    
+    resize_params.x = document.getElementById('resize_x_val').value;
+    resize_params.y = document.getElementById('resize_y_val').value;
+    resize_params.width = document.getElementById('resize_width').value;
+    resize_params.height = document.getElementById('resize_height').value;
+
+    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_RESIZE, resize_params);
```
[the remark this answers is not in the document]
juliareichel: fixed in new PR

examples/player/simid_player.html
ryanthompson591: Julia and I discussed this offline and I will leave it up to her.

PR #373, opened 2020-07-27
X button bug

juliareichel:
-Fixed bug where iframe listeners weren't being cleared properly so the iframe was considered null when it was actually initialized --> this problem created the error where you were unable to close and iframe, initialize a new one and close the second iframe because the functions would still try to close the first iframe the second time around
-Fixed the bug where the player buttons didn't disable when closing the ad from the creative

### DIFF
--- a/examples/player/simid_player.html
+++ b/examples/player/simid_player.html
@@ -23,9 +23,9 @@
     }
   </style>
   <script>
-    /** DISABLED and ENABLED constants are used to control the state of buttons. */
-    const DISABLED = true;
-    const ENABLED = false;
+    // /** DISABLED and ENABLED constants are used to control the state of buttons. */
+    // const DISABLED = true;
+    // const ENABLED = false;
 
     //if ad is non-linear allow user to insert specific dimensions
     function enableDisable() {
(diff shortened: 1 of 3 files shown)


examples/player/simid_player.js
```diff
@@ -477,6 +495,7 @@ class SimidPlayer {
       const closeMessage = {
         'code': reason,
       }
+      this.setCreativeControlsState_(/* controlState= */ DISABLED);
```
ryanthompson591, line 498: Could this be missed if there were a skip ad or fatal error?  How about putting this code in destroyIframeAndResumeContent_

juliareichel: Well in our sample creative we don't show demonstrate a skip ad or fatal error functionality via the creative (would seem counter intuitive if the creative skipped itself). But if you want me to move it for versatility then I can

examples/player/simid_player.html
```diff
@@ -79,7 +64,7 @@
       }
       simidPlayer.stopAd();
       simidPlayer.sendLog("User clicked close ad button on player");
-      setCreativeControlsState(/* controlState= */ DISABLED);
+      simidPlayer.setCreativeControlsState_(/* controlState= */ DISABLED);
```
ryanthompson591, line 67: probably also fine to remove the controlState comment now.

juliareichel: i spoke to Dan about this, since we are using consts that are really just booleans it is good to explain what the parameters actually are

damargulis: That was more since you were originally just using the literal `true` and `false`.  Now that you've named them, you can remove them if you want (also doesn't hurt to leave it in if you think it adds additional value)

juliareichel: Since this function is being called in this file as well as the nonlinear_tester and I think it definitely adds value there since the function isn't defined in that file, then we should stay consistent and it doesn't hurt

ryanthompson591: Using comments is common when you use a primitive type like:

setControlState(false);

A reader would wonder what false meant. So you would add a comment.

However, setControlState(ENABLED); is perfectly readable and understandable on its own.

In cases where you do need to define what a variable is, prefer putting that right into it's name. In this case it would be
const CONTROL_STATE_ENABLED = true;

Your function would then become.
setControlState(CONTROL_STATE_ENABLED);  This would also be readable, but IMO overly verbose.  

And this is right out.
setControlState( /*control state */ CONTROL_STATE_ENABLED);

Hopefully that makes sense, I wish I could link effective java but there essentially a whole chapter on using enums or constants as a way of making code clear and readable without excessive comments.

PR #378, opened 2020-07-29
Duration change bug

juliareichel:
-fixed bug that didn't allow a request duration to shorten the ad
-refactored request duration logic to deal with shortening the ad, extending the ad, and what happens if the creative requests a negative duration

### DIFF
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -20,16 +20,28 @@ class Extender extends BaseSimidCreative {
   /** @override */
   onGetMediaStateResolve(data) {
     super.onGetMediaStateResolve(data);
+    console.log(data);
     const mediaDuration = data['duration'];
+    console.log("media duration: " + mediaDuration);
     const params = {
-      'duration': mediaDuration + 5,
+      'duration': mediaDuration + 4,
     };
+
+    if (params.duration < 0) {
+      console.log('Requested duration too short. Requested duration: ' + params.duration);
+    }
+
     // Ask the player if we can extend duration 5 seconds to show end cards.
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
       this.extendDuration_ = true;
     }).catch(() => {
-      console.log('Player does not support extended duration.');
+      console.log('Player does not support requested duration change.');
     });
+  }
+
+  /** @override */
+  onTimeUpdate(data) {
+    console.log(data.args.currentTime);
   }
 
   onVideoEnded() {
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -20,11 +20,9 @@ class Extender extends BaseSimidCreative {
   /** @override */
   onGetMediaStateResolve(data) {
     super.onGetMediaStateResolve(data);
-    console.log(data);
     const mediaDuration = data['duration'];
-    console.log("media duration: " + mediaDuration);
     const params = {
-      'duration': mediaDuration + 5,
+      'duration': mediaDuration - 35,
     };
 
     if (params.duration < 0) {
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -24,7 +24,7 @@ class Extender extends BaseSimidCreative {
     const mediaDuration = data['duration'];
     console.log("media duration: " + mediaDuration);
     const params = {
-      'duration': mediaDuration - 31,
+      'duration': mediaDuration + 5,
     };
 
     if (params.duration < 0) {
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -24,7 +24,7 @@ class Extender extends BaseSimidCreative {
     const mediaDuration = data['duration'];
     console.log("media duration: " + mediaDuration);
     const params = {
-      'duration': mediaDuration + 4,
+      'duration': mediaDuration - 31,
     };
 
     if (params.duration < 0) {
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -22,7 +22,7 @@ class Extender extends BaseSimidCreative {
     super.onGetMediaStateResolve(data);
     const mediaDuration = data['duration'];
     const params = {
-      'duration': mediaDuration - 35,
+      'duration': mediaDuration + 5,
     };
 
     // Ask the player if we can extend duration 5 seconds to show end cards.
--- a/examples/creatives/extender.js
+++ b/examples/creatives/extender.js
@@ -25,21 +25,12 @@ class Extender extends BaseSimidCreative {
       'duration': mediaDuration - 35,
     };
 
-    if (params.duration < 0) {
-      console.log('Requested duration too short. Requested duration: ' + params.duration);
-    }
-
     // Ask the player if we can extend duration 5 seconds to show end cards.
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
       this.extendDuration_ = true;
     }).catch(() => {
       console.log('Player does not support requested duration change.');
     });
-  }
-
-  /** @override */
-  onTimeUpdate(data) {
-    console.log(data.args.currentTime);
   }
 
   onVideoEnded() {
--- a/examples/creatives/simid_overlay.js
+++ b/examples/creatives/simid_overlay.js
@@ -65,9 +65,7 @@ class SimidOverlay extends BaseSimidCreative {
       const params = {
         'duration': durationInput,
       };
-      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
-        this.extendDuration_ = true;
-      }).catch(() => {
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).catch(() => {
         console.log('Player does not support requested duration change.');
       });
     }
--- a/examples/creatives/simid_overlay.js
+++ b/examples/creatives/simid_overlay.js
@@ -27,7 +27,7 @@ class SimidOverlay extends BaseSimidCreative {
     this.sendMessageOnButtonClick_('request_skip', CreativeMessage.REQUEST_SKIP);
     this.sendMessageOnButtonClick_('request_stop', CreativeMessage.REQUEST_STOP);
     this.sendMessageOnLog_();
-    this.sendChangeAdDuration_();
+    this.sendMessageOnChangeDurationClick_();
   }
 
   /**
@@ -59,7 +59,7 @@ class SimidOverlay extends BaseSimidCreative {
       'click', sendLogFunction.bind(this));
   }
 
-  sendChangeAdDuration_() {
+  sendMessageOnChangeDurationClick_() {
     const sendAdDurationFunction = () => {
       const durationInput = document.getElementById('change_duration_input').value;
       const params = {
--- a/examples/creatives/simid_overlay.js
+++ b/examples/creatives/simid_overlay.js
@@ -27,6 +27,7 @@ class SimidOverlay extends BaseSimidCreative {
     this.sendMessageOnButtonClick_('request_skip', CreativeMessage.REQUEST_SKIP);
     this.sendMessageOnButtonClick_('request_stop', CreativeMessage.REQUEST_STOP);
     this.sendMessageOnLog_();
+    this.sendChangeAdDuration_();
   }
 
   /**
@@ -56,5 +57,21 @@ class SimidOverlay extends BaseSimidCreative {
     }
     document.getElementById('log_button').addEventListener(
       'click', sendLogFunction.bind(this));
+  }
+
+  sendChangeAdDuration_() {
+    const sendAdDurationFunction = () => {
+      const durationInput = document.getElementById('change_duration_input').value;
+      const params = {
+        'duration': durationInput,
+      };
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
+        this.extendDuration_ = true;
+      }).catch(() => {
+        console.log('Player does not support requested duration change.');
+      });
+    }
+    document.getElementById('change_duration_button').addEventListener(
+      'click', sendAdDurationFunction.bind(this));
   }
 }
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -436,10 +436,7 @@ class SimidPlayer {
     this.videoTrackingEvents_.set("durationchange", () => {
       this.simidProtocol.sendMessage(MediaMessage.DURATION_CHANGED);
     });
-    // this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
-    this.videoTrackingEvents_.set("ended", () => {
-      this.simidProtocol.sendMessage(MediaMessage.ENDED);
-    });
+    this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
     this.videoTrackingEvents_.set("error", () => {
       this.simidProtocol.sendMessage(MediaMessage.ERROR,
         {
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -469,6 +469,7 @@ class SimidPlayer {
     this.videoTrackingEvents_.set("timeupdate", () => {
       this.simidProtocol.sendMessage(MediaMessage.TIME_UPDATE,
         {'currentTime': this.adVideoElement_.currentTime});
+        this.compareAdAndRequestedDurations();
     });
     this.videoTrackingEvents_.set("volumechange", () => {
       this.simidProtocol.sendMessage(MediaMessage.VOLUME_CHANGE,
@@ -643,25 +644,12 @@ class SimidPlayer {
       }
       this.requestedDuration_ = originalDuration;
       this.simidProtocol.reject(incomingMessage, durationErrorMessage);
-    } else if (this.requestedDuration_ == UNLIMITED_DURATION) {
-      //Note: Users can end the ad using the close ad button on the player
-      return;
-    } else {
+    }
+    else {
       //If requested duration is any other acceptable value
-      this.changeAdEndTime();
+      this.compareAdAndRequestedDurations();
       this.simidProtocol.resolve(incomingMessage);
     }
-  }
-
-  /**
-   * Compares the requested change duration with the current ad time
-   *  to determine when to stop the ad, every 250 milliseconds.
-   * @private
-   */
-  changeAdEndTime() {
-    this.durationInterval_ = setInterval(() => {
-        this.compareAdAndRequestedDurations(); 
-    }, INTERVAL_TIMER);
   }
 
   /**
@@ -670,10 +658,15 @@ class SimidPlayer {
    * @private
    */
   compareAdAndRequestedDurations() {
-    if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
+    if (this.requestedDuration_ == NO_REQUESTED_DURATION || 
+      this.requestedDuration_ == UNLIMITED_DURATION) {
+        //Note: Users can end the ad with unlimited duration with
+        // the close ad button on the player
+        return;
+      }
+    else if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
       //Creative requested a duration shorter than the ad
       this.stopAd(StopCode.CREATIVE_INITATED);
-      clearInterval(this.durationInterval_);
       return;
     }
   } 
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -616,16 +616,16 @@ class SimidPlayer {
   videoComplete() {
     this.simidProtocol.sendMessage(MediaMessage.ENDED);
 
+    if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
+      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+    }
+
     //If the request duration is longer than the ad duration, the ad extends for the requested amount of time
     if (this.requestedDuration_ != UNLIMITED_DURATION) {
       const durationChangeMs = (this.requestedDuration_ - this.adVideoElement_.duration) * 1000;
       setTimeout(() => {
         this.stopAd(StopCode.CREATIVE_INITIATED);
       }, durationChangeMs);
-    }
-
-    if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
-      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
     }
   }
 
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -631,15 +631,13 @@ class SimidPlayer {
   onRequestChangeAdDuration(incomingMessage) {
     this.requestedDuration_  = incomingMessage.args['duration'];
     //If requested duration is negative seconds, play entire ad
-    if (this.requestedDuration_ < 0) {
+    if (this.requestedDuration_ != UNLIMITED_DURATION && this.requestedDuration_< 0) {
       this.sendLog("requested duration too short: " + this.requestedDuration_ + " seconds");
       setTimeout(() => {
         this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+        this.simidProtocol.reject(incomingMessage);
       }, this.adVideoElement_.duration * 1000);
-    } //If requested duration is 0 seconds, stop ad immediately  
-    else if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
-      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
-    } /*If requested duration is positive amount of seconds, call
+    }/*If requested duration is positive amount of seconds, call
       changeAdEndTime()*/
     else {
       this.changeAdEndTime();
--- a/examples/player/simid_player.js
+++ b/examples/player/simid_player.js
@@ -468,26 +468,30 @@ class SimidPlayer {
   }
 
   /**
-   * Called when creative requests a change in duration of ad.
-   * @param {int} requestedDuration The duration change request made
-   *   by the creative.
+   * Compares the requested change duration with the current ad time
+   *  to determine when to stop the ad, every 250 milliseconds.
    * @private
    */
-  changeAdEndTime(requestedDuration) {
+  changeAdEndTime() {
     this.interval_ = setInterval(() => {
-        this.compareCurrentTimeAndRequestedDuration(requestedDuration); 
+        this.compareAdAndRequestedDurations(); 
     }, 250);
   }
 
-  compareCurrentTimeAndRequestedDuration(requestedDuration) {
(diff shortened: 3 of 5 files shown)


examples/creatives/extender.js
```diff
@@ -22,13 +22,14 @@ class Extender extends BaseSimidCreative {
     super.onGetMediaStateResolve(data);
     const mediaDuration = data['duration'];
     const params = {
-      'duration': mediaDuration + 5,
+      'duration': mediaDuration - 35,
```
ryanthompson591, line 25: I think you want to revert this, since you were just doing this for testing.

examples/player/simid_player.js
```diff
@@ -431,7 +430,7 @@ class SimidPlayer {
     this.videoTrackingEvents_.set("durationchange", () => {
       this.simidProtocol.sendMessage(MediaMessage.DURATION_CHANGED);
     });
-    this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
+    // this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
```
ryanthompson591, line 433: don't we need to keep this callback?

kristen-m, line 433: Remove commented out code

examples/player/simid_player.js
```diff
@@ -609,14 +624,27 @@ class SimidPlayer {
...
+  /**
+   * Called when creative requests a change in duration of ad.
+   * @private
+   */
   onRequestChangeAdDuration(incomingMessage) {
-    if (this.requestedDuration_ != NO_REQUESTED_DURATION) {
-      // TODO: Support multiple change duration requests.
-      this.simidProtocol.reject(incomingMessage);
+    this.requestedDuration_  = incomingMessage.args['duration'];
+    //If requested duration is negative seconds, play entire ad
```
ryanthompson591, line 633: There is actually a negative number that can be passed in to requested duration.

https://interactiveadvertisingbureau.github.io/SIMID/#simid-creative-requestChangeAdDuration

One way to do this is to allow -2 and reject anything else. The other would be to stop the ad with any other negative values.

examples/player/simid_player.js
```diff
@@ -609,14 +624,27 @@ class SimidPlayer {
...
+    if (this.requestedDuration_ < 0) {
+      this.sendLog("requested duration too short: " + this.requestedDuration_ + " seconds");
+      setTimeout(() => {
+        this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+      }, this.adVideoElement_.duration * 1000);
+    } //If requested duration is 0 seconds, stop ad immediately
```
ryanthompson591, line 639: you probably don't even need to make any special code for this. If you set a low value (like 0) the rest of your code will probably just make this end without a lot of extra code.

examples/player/simid_player.js
```diff
@@ -609,14 +624,27 @@ class SimidPlayer {
...
+    this.requestedDuration_  = incomingMessage.args['duration'];
+    //If requested duration is negative seconds, play entire ad
+    if (this.requestedDuration_ < 0) {
+      this.sendLog("requested duration too short: " + this.requestedDuration_ + " seconds");
+      setTimeout(() => {
+        this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+      }, this.adVideoElement_.duration * 1000);
+    } //If requested duration is 0 seconds, stop ad immediately  
+    else if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
+      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
```
ryanthompson591, line 641: you need to always resolve or reject the requestChangeDuration message.

examples/player/simid_player.js
juliareichel: we no longer have a videoComplete function

examples/player/simid_player.js
juliareichel: good call

examples/player/simid_player.js
juliareichel: okay great, I altered the code to check for the -2 and rejected other values

examples/player/simid_player.js
juliareichel: fixed

examples/creatives/extender.js
juliareichel: yes

examples/player/simid_player.js
ryanthompson591: So if the creative does't call requestChangeDuration, how does the creative know when to terminate?

ryanthompson591: How do we know when to terminate if there is no new duration?

juliareichel: it does call requestChangeAdDuration and that function does exist

examples/player/simid_player.js
```diff
@@ -469,22 +468,36 @@ class SimidPlayer {
...
+   * Compares the requested change duration with the current ad time
+   *  to determine when to stop the ad, every 250 milliseconds.
    * @private
    */
-  videoComplete() {
-      this.simidProtocol.sendMessage(MediaMessage.ENDED);
+  changeAdEndTime() {
+    this.interval_ = setInterval(() => {
+        this.compareAdAndRequestedDurations(); 
+    }, 250);
```
Kaleab470, line 478: Use constant here

examples/creatives/extender.js
```diff
@@ -24,11 +24,12 @@ class Extender extends BaseSimidCreative {
     const params = {
       'duration': mediaDuration + 5,
     };
+
     // Ask the player if we can extend duration 5 seconds to show end cards.
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
       this.extendDuration_ = true;
     }).catch(() => {
-      console.log('Player does not support extended duration.');
+      console.log('Player does not support requested duration change.');
```
Kaleab470, line 32: Use sendLog here

juliareichel: sendLog is meant to send a message to the Player from the creative. It would not make sense to send a message to the player saying player does not support duration

examples/player/simid_player.js
juliareichel: fixed

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
   onRequestChangeAdDuration(incomingMessage) {
-    if (this.requestedDuration_ != NO_REQUESTED_DURATION) {
-      // TODO: Support multiple change duration requests.
-      this.simidProtocol.reject(incomingMessage);
+    this.requestedDuration_  = incomingMessage.args['duration'];
+    //If requested duration is negative seconds, play entire ad
+    if (this.requestedDuration_ != UNLIMITED_DURATION && this.requestedDuration_< 0) {
+      this.sendLog("requested duration too short: " + this.requestedDuration_ + " seconds");
+      setTimeout(() => {
+        this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
```
ryanthompson591, line 617: change stop code to creative requested complete (since the media wasn't complete).

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
+  /**
+   * Called when creative requests a change in duration of ad.
+   * @private
+   */
   onRequestChangeAdDuration(incomingMessage) {
-    if (this.requestedDuration_ != NO_REQUESTED_DURATION) {
-      // TODO: Support multiple change duration requests.
-      this.simidProtocol.reject(incomingMessage);
+    this.requestedDuration_  = incomingMessage.args['duration'];
+    //If requested duration is negative seconds, play entire ad
```
ryanthompson591, line 613: I suggest just removing this comment, because I don't think it's accurate anymore.

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
+    if (this.requestedDuration_ != UNLIMITED_DURATION && this.requestedDuration_< 0) {
+      this.sendLog("requested duration too short: " + this.requestedDuration_ + " seconds");
+      setTimeout(() => {
```
ryanthompson591, line 616: I don't think this is what we want.

How about,
if (this.requestedDuration_ != UNLIMITED_DURATION && this.requestedDuration_< 0) {
  const errorMessage = {
        errorCode : PlayerErrorCode:UNSUPPORTED_TIME,
        message: 'A negative duration is not valid.'
      }
      this.simidProtocol.reject(incomingMessage, errorMessage);
  }

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
+        this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+        this.simidProtocol.reject(incomingMessage);
+      }, this.adVideoElement_.duration * 1000);
+    }//If requested duration is positive amount of seconds, call changeAdEndTime()
```
ryanthompson591, line 620: move this comment into the else block.

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
+        this.stopAd(StopCode.CREATIVE_INITATED);
+        clearInterval(this.interval_);
+    } else if (this.requestedDuration_ >= this.adVideoElement_.duration) {
+      const durationChangeMs = (this.requestedDuration_ - this.adVideoElement_.duration) * 1000;
+      setTimeout(() => {
+        this.stopAd(StopCode.CREATIVE_INITIATED);
+      }, durationChangeMs);
+      clearInterval(this.interval_);
+    }
+    this.simidProtocol.sendMessage(MediaMessage.ENDED);
```
ryanthompson591, line 656: we can remove this because now we're sending it directly when we get the ended message from the video element.

examples/player/simid_player.js
```diff
@@ -609,14 +592,68 @@ class SimidPlayer {
...
+
+  /**
+   * Compares the duration of the ad with the requested change duration.
+   * If request duration is shorter, the ad stops early. 
+   * If the request duration is longer, the ad extends for the requested
+   *  amount of time.
+   * @private
+   */
+  compareAdAndRequestedDurations() {
+    if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
```
ryanthompson591, line 646: So what I was trying to explain earlier is that this logic is a little delicate.

For example, imagine this scenario:
currentTime is 10 seconds
videoTime is 20 seconds
RequestedTime is 30 seconds

The very first time we get into this section the statement this.adVideoElement_.currentTime >= this.requestedDuration_

will evaluate to false.

In javascript this means we will get into the else section
this.requestedDuration_ >= this.adVideoElement.duration

will evaluate to true.  This will then start the timer.

-----
Now imagine that the user pauses the ad with the pause button and makes a sandwich.  The ad will be terminated by the time they get back, which is not what we want. The desired effect is that unpausing should resume where they left off.

Instead consider this logic.

if (this.adVideoElement.ended) {
  // The video has ended but the creative has request more time, so set a timer to remove the creative for that case.
  clearInterval(this.interval_);
  if (this.requestedDuration_ == -2) {
   // The creative has asked for unlimited time, there will be no timeout. 
   // Note: We suggest any player allowing unlimited time to implement its own close ad button.  This will prevent bad user experiences.
  return;
  }
const durationChangeMs = (this.requestedDuration_ - this.adVideoElement_.duration) * 1000;
      setTimeout(() => {
        this.stopAd(StopCode.CREATIVE_INITIATED);
      }, durationChangeMs);
} else if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
   // The creative has requested a shorted duration than the video's duration. This player will now stop the ad.
    clearInterval(this.interval_);
    this.stopAd(StopCode.CREATIVE_INITATED);
}

examples/player/simid_player.js
juliareichel: the media is complete in this situation because when the creative requests negative duration I play the entire ad

examples/player/simid_player.js
juliareichel: okay deleted

examples/player/simid_player.js
juliareichel: okay, if we want this logic then I will change the stopAd code to creative requested

examples/player/simid_player.js
juliareichel: moved

examples/player/simid_player.js
juliareichel: okay

examples/creatives/simid_overlay.js
```diff
@@ -57,4 +58,20 @@ class SimidOverlay extends BaseSimidCreative {
     document.getElementById('log_button').addEventListener(
       'click', sendLogFunction.bind(this));
   }
+
+  sendChangeAdDuration_() {
```
ryanthompson591, line 62: This function reads like it will actually send a change Ad Duration call.

Instead what it is really doing is setting up the button listener.

I suggest the name:
sendMessageOnChangeDurationClick_

It's a little verbose but easy to understand.

examples/creatives/simid_overlay.js
```diff
@@ -57,4 +58,20 @@ class SimidOverlay extends BaseSimidCreative {
...
+    const sendAdDurationFunction = () => {
```
ryanthompson591, line 63: These sorts of anonymous function are fine, but most developers would probably prefer to make this a class level function.

I think the reason is that it makes debugging easier.  If there is a stack trace, these will show as anonymous function whereas using a class function will show line numbers.

examples/player/simid_player.js
```diff
@@ -431,7 +431,10 @@ class SimidPlayer {
     this.videoTrackingEvents_.set("durationchange", () => {
       this.simidProtocol.sendMessage(MediaMessage.DURATION_CHANGED);
     });
-    this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
+    // this.videoTrackingEvents_.set("ended", this.videoComplete.bind(this));
```
ryanthompson591, line 434: remove the commented code since you have it defined better below.

examples/player/simid_player.js
```diff
@@ -609,14 +595,82 @@ class SimidPlayer {
     console.log('The creative has asked for the player to ping ' + requestedUrlArray);
   }
 
+  /**
+   * Called when video playback is complete.
+   * @private
+   */
+  videoComplete() {
```
ryanthompson591, line 602: is this function called anymore?  Looks like you commented out the code that calls this.

examples/player/simid_player.js
```diff
@@ -609,14 +595,82 @@ class SimidPlayer {
...
-      this.simidProtocol.reject(incomingMessage);
+    this.requestedDuration_  = incomingMessage.args['duration'];
+    if (this.requestedDuration_ != UNLIMITED_DURATION && this.requestedDuration_< 0) {
+      const durationErrorMessage = {
+        errorCode: PlayerErrorCode.UNSUPPORTED_TIME,
+        message: 'A negative duration is not valid.'
+      }
+      this.simidProtocol.reject(incomingMessage, durationErrorMessage);
+    } else if (this.requestedDuration_ == UNLIMITED_DURATION) {
+      //Note: User can x out of the ad using the close ad button on the player
```
ryanthompson591, line 623: s/x out of the ad/end the ad

examples/player/simid_player.js
```diff
@@ -609,14 +595,82 @@ class SimidPlayer {
...
+      }, durationChangeMs);
+      clearInterval(this.interval_);
+      return;
+    } else if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
+      //Creative requested a duration shorter than the ad
+      this.stopAd(StopCode.CREATIVE_INITATED);
+      clearInterval(this.interval_);
+      return;
+    } else if (this.requestedDuration_ >= this.adVideoElement_.duration) {
+      //Creative requested a longer duration
```
ryanthompson591, line 666: Let's make this comment meaningful to a first time reader.

// Creative requested a duration longer than the video, this will cause the ad to extend after when the video completes.

examples/player/simid_player.js
```diff
@@ -609,14 +595,82 @@ class SimidPlayer {
...
+        this.stopAd(StopCode.CREATIVE_INITIATED);
+      }, durationChangeMs);
+      clearInterval(this.interval_);
+      return;
+    } else if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
+      //Creative requested a duration shorter than the ad
+      this.stopAd(StopCode.CREATIVE_INITATED);
+      clearInterval(this.interval_);
+      return;
+    } else if (this.requestedDuration_ >= this.adVideoElement_.duration) {
```
ryanthompson591, line 665: I think it may be possible to remove this entire else if section since you are doing exactly this if the video ends and the user requested duration is not yet reached.

examples/creatives/simid_overlay.js
juliareichel: changed name

examples/player/simid_player.js
juliareichel: changed

examples/player/simid_player.js
juliareichel: tested this and worked with a little refactoring

examples/player/simid_player.js
juliareichel: uncommented

examples/player/simid_player.js
juliareichel: changed back to original

examples/player/simid_player.js
juliareichel: understood

examples/player/simid_player.js
```diff
@@ -609,16 +602,78 @@ class SimidPlayer {
...
+
+  /**
+   * Compares the duration of the ad with the requested change duration.
+   * If request duration is shorter, the ad stops early. 
+   * If the request duration is longer, the ad extends for the requested
+   *  amount of time.
+   * @private
+   */
+  compareAdAndRequestedDurations() {
+    const durationChangeMs = (this.requestedDuration_ - this.adVideoElement_.duration) * 1000;
```
damargulis, line 660: this only gets used inside of the if clause, so define it in there

examples/player/simid_player.js
```diff
@@ -609,16 +602,78 @@ class SimidPlayer {
...
+
+    if (this.adVideoElement_.ended) {
+      //If the video ad has ended already close the creative at the requested duration
+      setTimeout(() => {
+        this.stopAd(StopCode.CREATIVE_INITIATED);
+      }, durationChangeMs);
+      clearInterval(this.interval_);
+      return;
```
damargulis, line 668: Seems odd to use a timer to check if the video element has ended, when it dispatches an ENDED event that we are already listening for.  Can this logic be moved to the videoComplete function?

examples/player/simid_player.js
```diff
@@ -609,16 +602,78 @@ class SimidPlayer {
...
-    this.simidProtocol.resolve(incomingMessage);
   }
 
+  /**
+   * Compares the requested change duration with the current ad time
+   *  to determine when to stop the ad, every 250 milliseconds.
+   * @private
+   */
+  changeAdEndTime() {
+    this.interval_ = setInterval(() => {
```
damargulis, line 647: Make sure to add `this.interval_` to the constructor (even if you're just setting it to null), and include type annotations

examples/creatives/simid_overlay.js
juliareichel: weird, it must have not been added properly. Just committed it again

examples/creatives/simid_overlay.js
```diff
@@ -57,4 +58,20 @@ class SimidOverlay extends BaseSimidCreative {
...
+      const durationInput = document.getElementById('change_duration_input').value;
+      const params = {
+        'duration': durationInput,
+      };
+      this.simidProtocol.sendMessage(CreativeMessage.REQUEST_CHANGE_AD_DURATION, params).then(() => {
+        this.extendDuration_ = true;
```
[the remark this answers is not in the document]
juliareichel: Included that accidentally from the extender.js file

examples/player/simid_player.js
juliareichel: Added!

examples/player/simid_player.js
juliareichel: @ryanthompson591 I moved the logic to the videoComplete and it didn't seem to cause any issues. Do you have a strong opinion on keeping this check within this function or do you think it can be moved?

juliareichel: spoke with Ryan and he said either is fine, so moved it

examples/player/simid_player.js
```diff
@@ -609,16 +609,75 @@ class SimidPlayer {
...
+    }, INTERVAL_TIMER);
+  }
+
+  /**
+   * Compares the duration of the ad with the requested change duration.
+   * If request duration is shorter, the ad stops early. 
+   * @private
+   */
+  compareAdAndRequestedDurations() {
+    if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
```
ryanthompson591, line 673: You might hate me, but I promise we're working towards better code. What do you think of just putting that into the currentTimeUpdate callback?

We wouldn't need this interval at all anymore.  You would have to add one more conditional there that we weren't NO_REQUESTED_DURATION or UNLIMITED_DURATION.  But you could probably delete a bunch of complexity and keep this code more simple.


Basically, you would modify the code on line 563 (of the current player). Look for this:
    this.videoTrackingEvents_.set("timeupdate", () => {

You would change it to this:
this.videoTrackingEvents_.set("timeupdate", () => {
      this.simidProtocol.sendMessage(MediaMessage.TIME_UPDATE,
        {'currentTime': this.adVideoElement_.currentTime});
       this.compareAdAndRequestedDuration();
    });

Then modify your compareAdAndRequestedDuration function with this:
compareAdAndRequestedDurations() {
   if (this.requestedDuration_ == NO_REQUESTED_DURATION ||
    this.requestedDuration_ == UNLIMITED_DURATION) {
   // We don't terminate the ad in these cases here.
    return;
}
    if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
///.... the same code you have...


Then you would get rid of the line:

this.changeAdEndTime();

and you would get rid of the function
changeAdEndTime

You wouldn't need those anymore, because any time any time update happened you would have a call to compareAdAndRequestedDurations

examples/player/simid_player.js
```diff
@@ -609,16 +609,75 @@ class SimidPlayer {
...
+    if (this.requestedDuration_ != UNLIMITED_DURATION) {
+      const durationChangeMs = (this.requestedDuration_ - this.adVideoElement_.duration) * 1000;
+      setTimeout(() => {
+        this.stopAd(StopCode.CREATIVE_INITIATED);
+      }, durationChangeMs);
+    }
+
+    if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
+      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+    }
```
damargulis, line 629: I think this segment should go above the other if block, and return so that if there is no request duration change, you don't ever even start the timer

juliareichel: moved

examples/player/simid_player.js
juliareichel: Happy to keep learning better coding practices. Just made the changes, let me know your thoughts.

examples/creatives/simid_overlay.js
juliareichel: Got it

examples/player/simid_player.js
```diff
@@ -609,16 +609,69 @@ class SimidPlayer {
...
+   */
+  videoComplete() {
+    this.simidProtocol.sendMessage(MediaMessage.ENDED);
+
+    if (this.requestedDuration_ == NO_REQUESTED_DURATION) {
+      this.stopAd(StopCode.MEDIA_PLAYBACK_COMPLETE);
+    }
+
+    //If the request duration is longer than the ad duration, the ad extends for the requested amount of time
+    if (this.requestedDuration_ != UNLIMITED_DURATION) {
```
damargulis, line 624: Should this be an `else if`? (or a return inside the if block above?)

Otherwise, if the requestedduration_ == NO_REQUESTED_DUR, then well stop the ad, but also come down here and start a timer to stop the ad again in a few seconds

examples/player/simid_player.js
```diff
@@ -609,16 +609,69 @@ class SimidPlayer {
...
+  /**
+   * Called when creative requests a change in duration of ad.
+   * @private
+   */
   onRequestChangeAdDuration(incomingMessage) {
-    if (this.requestedDuration_ != NO_REQUESTED_DURATION) {
-      // TODO: Support multiple change duration requests.
-      this.simidProtocol.reject(incomingMessage);
+    const originalDuration = this.requestedDuration_;
+    this.requestedDuration_  = incomingMessage.args['duration'];
```
damargulis, line 638: imo this would read a little easier if you only assigned the new duration to `this.requestedDuration_` in the else branch, rather than storing the original and resetting

examples/player/simid_player.js
```diff
@@ -609,16 +609,69 @@ class SimidPlayer {
...
+    if (this.requestedDuration_ == NO_REQUESTED_DURATION || 
+      this.requestedDuration_ == UNLIMITED_DURATION) {
+        //Note: Users can end the ad with unlimited duration with
+        // the close ad button on the player
+        return;
+      }
+    else if (this.adVideoElement_.currentTime >= this.requestedDuration_) {
+      //Creative requested a duration shorter than the ad
+      this.stopAd(StopCode.CREATIVE_INITATED);
+      return;
```
damargulis, line 671: You don't need both the returns and the else/if for this function, either one or the other should suffice

examples/player/simid_player.js
```diff
@@ -609,16 +609,69 @@ class SimidPlayer {
...
   }
 
+  /**
+   * Compares the duration of the ad with the requested change duration.
+   * If request duration is the same as the ad duration, ad ends as normal.
+   * If request duration is unlimited, ad stays on screen until user closes ad.
+   * If request duration is shorter, the ad stops early. 
+   * @private
+   */
+  compareAdAndRequestedDurations() {
```
damargulis, line 661: Make sure any private functions names end with an underscore `_`.

juliareichel: added the _

examples/player/simid_player.js
juliareichel: yes, will change to else if

examples/player/simid_player.js
juliareichel: deleted return

examples/player/simid_player.js
juliareichel: if I do this, then in the case where the user inputs a negative value, the ad will stop immediately and not show the entire ad (which is the desired functionality)

damargulis: Can you explain why that would be the case?

if you change this so that its something like

```
const newDuration = incomingMessage.args['duration'];
if (newDuration != UNLIMITED && newDuration < 0) {
  this.simidProtocol.reject(...);
} else {
  this.requestedDuration_ = newDuration;
  ...
}
```

I don't think I see how the end result would be any different

juliareichel: Oh I misunderstood what you were referring to when you said new duration. I thought you meant to still use the this.reqeuestedDuration_ in the if statement which wouldn't work. This will work and I will change it

PR #380, opened 2020-07-29
Resize persist after expand/collapse

kristen-m:
Fixes bug that had resized nonlinear returning to original dimensions after expand/collapse instead of the dimensions it was resized to.


examples/player/simid_player.js
```diff
@@ -99,6 +99,18 @@ class SimidPlayer {
      */
     this.rejectInitializationPromise_ = null;
 
+    /**
+     * A boolean value that tracks whether or not the creative has been resized.
+     * @private {boolean}
+     */
+    this.isResized_ = false;
```
damargulis, line 106: Since resizeDimensions_ starts as null, I don't think you need this extra tracker variable

if (this.resizeDimensions_) should be enough to check if its been resized, since that will evaluate to false when resizeDimensions is null

examples/player/simid_player.js
```diff
@@ -287,8 +299,11 @@ class SimidPlayer {
     newDimensions.y = document.getElementById('y_val').value;
     newDimensions.width = document.getElementById('width').value;
     newDimensions.height = document.getElementById('height').value;
-
-    return newDimensions;
+    if(this.isResized_) {
```
damargulis, line 302: rework this a bit so that newDimensions isn't even created if it isn't needed

examples/player/simid_player.js
```diff
@@ -282,13 +288,16 @@ class SimidPlayer {
...
-    let newDimensions = {};
-    newDimensions.x = document.getElementById('x_val').value;
-    newDimensions.y = document.getElementById('y_val').value;
-    newDimensions.width = document.getElementById('width').value;
-    newDimensions.height = document.getElementById('height').value;
-
-    return newDimensions;
+    if(this.resizeDimensions_) {
+      return this.resizeDimensions_;
+    } else {
```
damargulis, line 293: Since you're returning up there, you don't even need to put this part in an else clause

examples/player/simid_player.js
```diff
@@ -99,6 +99,12 @@ class SimidPlayer {
      */
     this.rejectInitializationPromise_ = null;
 
+    /**
+     * An array of the resized nonlinear creative's dimensions.
```
ryanthompson591, line 103: Nit: this is an object not array.

examples/player/simid_player.js
```diff
@@ -282,13 +288,16 @@ class SimidPlayer {
...
    */
   getNonlinearDimensions_() {
-    let newDimensions = {};
-    newDimensions.x = document.getElementById('x_val').value;
-    newDimensions.y = document.getElementById('y_val').value;
-    newDimensions.width = document.getElementById('width').value;
-    newDimensions.height = document.getElementById('height').value;
-
-    return newDimensions;
+    if(this.resizeDimensions_) {
```
Kaleab470, line 291: nit - imo,  if this.resizeDimensions reads a little weird here

kristen-m: What do you think would sound better here?

examples/player/simid_player.js
kristen-m: Good catch, fixed

examples/player/simid_player.js
kristen-m: Fixed

examples/player/simid_player.js
Kaleab470: if you rename the variable to nonLinearDimensions_ it might help with readability here and also at line 407/8

ryanthompson591: hrm, yeah that gave my brain an idea. Why not just call it creativeDimensions_

When you initialize a nonlinear ad set creative dimensions to whatever is in the GUI.  maybe use a function getCreativeDimensionsFromUi().

Then anywhere else, just use this.creativeDimensions_

Would that read easier? Let me know what you think.

kristen-m: If we call it creativeDimensions_ do you think it would get confused with the linear creative dimensions? I might prefer nonLinearDimensions_ because it makes it clear the type of creative whose dimensions we want to keep track of.

PR #383, opened 2020-07-29
Add dimension tracking to base creative

kristen-m:
--Update the base SIMID creative so that it keeps track of both the creative dimensions and the player dimensions
--Update the nonlinear creative to make proper resize function call based off of updates

### DIFF
--- a/examples/creatives/base_simid_creative.js
+++ b/examples/creatives/base_simid_creative.js
@@ -248,7 +248,8 @@ class BaseSimidCreative {
     console.log("Received message from player: " + logMessage);
   }
 
-  /**Sends message requesting a to resize creative based off of given resizeParameters
+  /**
+   * Sends message requesting to resize creative based off of given resizeParameters
    * @param {!Object} resizeParams An object with the video & creative dimensions.
    */
   requestResize(resizeParams) {
--- a/examples/creatives/base_simid_creative.js
+++ b/examples/creatives/base_simid_creative.js
@@ -244,7 +244,7 @@ class BaseSimidCreative {
   }
 
   /**Sends message requesting a to resize creative based off of given resizeParameters
-   * @param {!Object} resizeParams
+   * @param {!Object} resizeParams an object with the video & creative dimensions.
    */
   requestResize(resizeParams) {
     this.simidProtocol.sendMessage(CreativeMessage.REQUEST_RESIZE, resizeParams).then( () => {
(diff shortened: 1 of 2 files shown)


examples/creatives/testers_nonlinear.js
```diff
@@ -40,10 +40,10 @@ class TestersNonLinear extends BaseSimidCreative {
     creativeDimensions.height = document.getElementById('resize_height').value;
 
     const resizeParams = {
-      mediaDimensions: {},
+      videoDimensions: this.environmentData.videoDimensions,
       creativeDimensions: creativeDimensions
     };
 
-    this.simidProtocol.sendMessage(CreativeMessage.REQUEST_RESIZE, resizeParams);
+    super.requestResize(resizeParams);
```
damargulis, line 47: Since TestersNonlinear doesn't override requestResize, you don't need to use super here, you can use `this.requestResize`

examples/creatives/base_simid_creative.js
```diff
@@ -236,4 +247,14 @@ class BaseSimidCreative {
     const logMessage = data.args.message;
     console.log("Received message from player: " + logMessage);
   }
+
+  /**Sends message requesting a to resize creative based off of given resizeParameters
```
damargulis, line 251: nit. fix the formatting here

juliareichel, line 251: * Sends message requesting TO (get rid of a)

examples/creatives/base_simid_creative.js
```diff
@@ -138,6 +139,16 @@ class BaseSimidCreative {
...
 
+  /**
+   * Called when the creative receives a resize message from the player.
+   * @param {!Object} eventData Data from the event.
+   * @protected
+   */
+  onReceiveResize(eventData) {
+    this.environmentData.creativeDimensions = eventData.args.creativeDimensions;
+    this.environmentData.videoDimensions = eventData.args.videoDimensions;
+  }
```
Kaleab470, line 150: would you want to resolve here?

ryanthompson591: In this case you wouldn't need to because resize is information only.
https://interactiveadvertisingbureau.github.io/SIMID/#reference-table

examples/creatives/base_simid_creative.js
kristen-m: done

examples/creatives/testers_nonlinear.js
kristen-m: fixed

examples/creatives/base_simid_creative.js
kristen-m: Good catch! Fixed

PR #385, opened 2020-07-30
Nonlinear duration

Kaleab470:


examples/player/simid_player.html
```diff
@@ -37,14 +37,17 @@
       let linearButton = document.getElementById("linear_ad");
       let creativeDimensions = document.getElementById("creative_dimensions");
       let mediaFile = document.getElementById("media_file");
+      let creativeduration = document.getElementById("creative_duration");
```
juliareichel, line 40: camelCase

examples/player/simid_player.js
```diff
@@ -306,6 +306,12 @@ class SimidPlayer {
       this.simidIframe_.style.position = "absolute";
 
       this.contentVideoElement_.play();
+
+      const nonlinearDuration = document.getElementById('duration').value * 1000;
```
juliareichel, line 310: * nonLinearDuration

examples/player/simid_player.js
```diff
@@ -430,6 +436,7 @@ class SimidPlayer {
 
     const creativeData = {
       'adParameters' : document.getElementById('ad_params').value,
+      'duration' : document.getElementById('duration').value,
```
damargulis, line 439: I think you only want to add this for nonlinear ads?

examples/player/simid_player.js
```diff
@@ -306,6 +306,12 @@ class SimidPlayer {
       this.simidIframe_.style.position = "absolute";
 
       this.contentVideoElement_.play();
+
+      const nonlinearDuration = document.getElementById('duration').value * 1000;
+      setTimeout(() => {
```
ryanthompson591, line 311: no let's not use setTimeout.  It's possible the user pauses playback (like when they go to a different tab or whatever).  Instead let's use currentTime.

So add some code to the currentTime callback that just does a comparison.

If (this.nonlinearStarTime_ - this.vidoeElement_.currentTime > duration) {
  this.stopAd(StopCode.UNSPECIFIED);
}


Side note: I think we need another stop code. NON_LINEAR_DURATION_COMPLETE

https://github.com/InteractiveAdvertisingBureau/SIMID/issues/389

examples/player/simid_player.html
Kaleab470: done

examples/player/simid_player.js
Kaleab470: good catch

examples/player/simid_player.js
Kaleab470: Done!

examples/player/simid_player.js
```diff
@@ -237,10 +250,15 @@ class SimidPlayer {
...
+    for(let [key, func] of this.adVideoTrackingEvents_) {
       this.adVideoElement_.removeEventListener(key, func, true);
     }
-    this.videoTrackingEvents_.clear();
+
+    for(let [key, func] of this.contentVideoTrackingEvents_) {
+      this.contentVideoElement_.removeEventListener(key, func, true);
+    }
+
+    this.adVideoTrackingEvents_.clear();
```
ryanthompson591, line 261: add
this.contentVideoElement_.clear()

Kaleab470: Good catch!

PR #391, opened 2020-08-03
Auto populate fields

Kaleab470:


examples/player/simid_player.html
```diff
@@ -145,6 +154,13 @@
           
           adParams = JSON.stringify(adParamsObj);
           break;
+        case 'testers_nonlinear.html':
+          document.getElementById("nonlinear_ad").click();
```
damargulis, line 158: if you change the listeners from onclick to onchange, then use this line to set the new value, instead of calling "click", does it still work? Imo that would be a bit better than triggering a fake click

Kaleab470: onchange doesn't get triggered if we set the value to be checked with javascript so we'd have to simulate a click

examples/player/simid_player.html
```diff
@@ -123,8 +123,10 @@
       let adParamsObj = '';
       switch (value) {
         case 'simid_overlay.html':
+          document.getElementById("linear_ad").click();
```
ryanthompson591, line 126: It's really strange to use click here, but I guess it's fine.

A more common way to do this might be:
  setAdTypeLinear();

Then have a function
function setAdTypeLinear() {
  document.getElementById("linear_ad").checked;
  setInputsForAdType();
}

Either way seems fine though.

examples/player/simid_player.html
juliareichel: Agreed, clicking seems a little weird

examples/player/simid_player.html
Kaleab470: I like this way better actually!

juliareichel:
This looks good- so should I integrate this methodology for the new creatives that I implemented?

PR #394, opened 2020-08-04
Hover

juliareichel:
-Created a banner to showcase the creative's ability to expand on hover and collapse on mouse out


examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,103 @@
...
+     * Adds a hover event listener to the contents of the iframe that expands the iframe.
+     * @param {String} elementName The name of the element.
+     * @param {Event} event The event performed on the element.
+     * @private
+     */   
+    onHover_(elementName, event) {
+        const expandOnHoverFunction = () => {
+            const newDimensions = {};
+
+            newDimensions.x = this.initialDimensions_.x * X_OFFSET_PERCENTAGE;
```
ryanthompson591, line 72: This logic will work for our player, but not necessarily on other players that might give different original dimensions right?

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,103 @@
...
+     * Adds a mouse out event listener to the contents of the iframe that shrinks the iframe
+     *  back to its original size.
+     * @param {String} elementName The name of the element.
+     * @param {Event} event The event performed on the element.
+     * @private
+     */  
+    onMouseOut_(elementName, event) {
+        const collpaseOnMouseOutFunction = () => {
+            const restoreParams = {
+                creativeDimensions: this.initialDimensions_,
```
ryanthompson591, line 96: add the video dimensions to this.  Should be something like
onst restoreParams = {
                creativeDimensions: this.initialDimensions_,
   videoDimensions:environmentData.videoDimensions,
}

examples/player/simid_player.html
```diff
@@ -200,6 +200,7 @@
         <button onclick="switchCreative('extender.html')">extender</button>
         <button onclick="switchCreative('testers_nonlinear.html')">testers_nonlinear</button>
         <button onclick="switchCreative('banner_nonlinear.html')">banner_nonlinear</button>
+        <button onclick="switchCreative('hover_nonlinear.html')">hover_nonlinear</button>
```
ryanthompson591, line 203: it would be cool if you added the defaults to switchCreative for this one.  Like switching the radio buttons to nonlinear and adding some default dimensions that demonstrate this correctly.

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,103 @@
...
+        const expandOnHoverFunction = () => {
+            const newDimensions = {};
+
+            newDimensions.x = this.initialDimensions_.x * X_OFFSET_PERCENTAGE;
+            newDimensions.y = this.initialDimensions_.y * Y_OFFSET_PERCENTAGE;
+            newDimensions.width = this.initialDimensions_.width * WIDTH_PERCENTAGE;
+            newDimensions.height = this.initialDimensions_.height * HEIGHT_PERCENTAGE;
+        
+            const resizeParams = {
+                creativeDimensions: newDimensions,
```
kristen-m, line 78: Do you need to include the videoDimensions in this parameter too?

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,103 @@
...
+    storeCreativeDimensions_() {
+        const creativeDimensions = {};
+
+        creativeDimensions.x = this.environmentData.creativeDimensions.x;
+        creativeDimensions.y = this.environmentData.creativeDimensions.y;
+        creativeDimensions.width = this.environmentData.creativeDimensions.width;
+        creativeDimensions.height = this.environmentData.creativeDimensions.height;
+
+        return creativeDimensions;
+    }
```
kristen-m, line 60: Instead of creating this function to store the creativeDimensions, would it be possible to simply set this.initialDimensions equal to this.environmentData.creativeData onInit?

examples/creatives/hover_nonlinear.js
juliareichel: the initialDimensions should be the same because I get the values from environmentData but the offset percentage that I used was specific for our player

examples/creatives/hover_nonlinear.js
juliareichel: Added

examples/creatives/hover_nonlinear.js
juliareichel: yes - I added it here too

examples/creatives/hover_nonlinear.js
juliareichel: this.environmentData.creativeDimensions would work, good suggestion

examples/player/simid_player.html
juliareichel: the default dimensions do demonstrate this correctly. I can definitely try switching the radio button to nonlinear

examples/player/simid_player.html
```diff
@@ -116,7 +116,17 @@
     }
 
     function switchCreative(value) {
-      document.getElementById('creative_url').value = '../creatives/' + value;
+      const creative_url = document.getElementById('creative_url').value = '../creatives/' + value;
+
+      if (creative_url.includes("nonlinear")){
```
Kaleab470, line 121: this seems like its prone to a lot of things - one of which is the need to have the word 'nonlinear' in every nonlinear ad created.

examples/player/simid_player.html
```diff
@@ -116,7 +116,17 @@
...
+        document.getElementById("nonlinear_ad").checked = true;
+        document.getElementById("linear_ad").checked = false;
```
Kaleab470, line 123: i had them set up so that only one is selected at a time, so if you check one it should be unchecking the other one

examples/player/simid_player.html
```diff
@@ -116,7 +116,17 @@
...
+      }
+      else {
+        document.getElementById("linear_ad").checked = true;
+        document.getElementById("nonlinear_ad").checked = false;
```
Kaleab470, line 127: moreover, i don't think that this setup works in making the creative dimension and duration field pop up automatically since the function that does that is called onclick.

Kaleab470:
The PR i have open right now should hopefully help with the checking/unchecking thing you do in your code

examples/player/simid_player.html
juliareichel: correct- this is just for our example player so I don't think it should be an issue. If you have any other suggestions let me know!

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,89 @@
+const Y_OFFSET_PERCENTAGE = 8;
+const X_OFFSET_PERCENTAGE = 8;
+const WIDTH_PERCENTAGE = 1.1;
+const HEIGHT_PERCENTAGE = 2;
+
+class HoverNonLinear extends BaseSimidCreative {
```
ryanthompson591, line 6: Add a description to what this sample is and what it does.

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,89 @@
...
+        super.onStart(eventData);
+        this.addActions_();
+    }
+
+    /**
+     * Adds actions to different buttons available on the overlay.
+     * @private
+     */
+    addActions_() {
+        this.sendMessageOnClick_("close_ad", CreativeMessage.REQUEST_STOP);
```
ryanthompson591, line 30: use single quotes everywhere instead of double quotes.

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,89 @@
...
+        this.onHover_("content_container", 'mouseover');
+        this.onMouseOut_("content_container", 'mouseout');
+    }
+
+    /**
+     * Listens for an event on the banner
```
ryanthompson591, line 36: Was this what my doc for this function was. Let's clean it up a bit.

Sends a SIMID message whenever an element is clicked.

examples/player/simid_player.html
Kaleab470: You wont be needing this piece of code in your PR anyway. I’ll merge my autopopulate PR which has this functionality so you just have to pull from master before you merge

juliareichel: yup, I commented on your PR

examples/player/simid_player.html
juliareichel: N/A after merge

examples/player/simid_player.html
juliareichel: N/A after merge

examples/creatives/hover_nonlinear.js
ryanthompson591: OK, I guess first these aren't really percentages, it's a little misleading.

For example, if y offset is 5 pixels and you multiply that by 8, the whole thing shifts down to 40 pixels.  Is that what we want?

How does this even work right? I think you need to demo this to me and we need to talk about how it should work.

The logic might be a little complex but I think we can get it to work right regardless of what the original dimensions are.

examples/player/simid_player.html
ryanthompson591: Yeah, prefer just manually setting each one, to a trick like this (though I do like the idea).

examples/creatives/hover_nonlinear.js
juliareichel: added

examples/creatives/hover_nonlinear.js
juliareichel: okay, will do

examples/creatives/hover_nonlinear.js
juliareichel: will do!

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+
+        this.initialDimensions_ = null;
+    }
+
+    /** @override */
+    onInit(eventData) {
+        super.onInit(eventData);
+        this.initialDimensions_ = this.environmentData.creativeDimensions;
+        this.videoDimensionHeight = this.environmentData.videoDimensions.height;
+        this.videoDimensionWidth = this.environmentData.videoDimensions.width;
```
Kaleab470, line 19: you are declaring private instance variables here so i'd suggest creating them first in the constructor and updating them here as it isn't very clear that they persisit

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+        const expandOnHoverFunction = () => {
+            const newDimensions = {};
+            let resizeParams = {};
+
+            //if ad is at the top quarter of the player
+            if (this.initialDimensions_.y < (this.videoDimensionHeight * QUARTER_THE_SIZE)) {
+                newDimensions.height = (this.videoDimensionHeight * THIRTY_FIVE_PERCENT);
+                newDimensions.x = 0;
+                newDimensions.y = this.initialDimensions_.y;
+                newDimensions.width = this.videoDimensionWidth;
```
Kaleab470, line 65: looks like newDimensions.x and newDimensions.width are always set to the same value no matter which conditional gets run. so take those two lines and put them right after you declare newDimensions

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Adds a hover event listener to the contents of the iframe that expands the iframe.
+     * @param {String} elementName The name of the element.
+     * @param {Event} event The event performed on the element.
+     * @private
+     */   
+    onHover_(elementName, event) {
+        const expandOnHoverFunction = () => {
+            const newDimensions = {};
+            let resizeParams = {};
```
Kaleab470, line 58: resizeParams should also be a const here

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+                newDimensions.x = 0;
+                newDimensions.y = this.initialDimensions_.y;
+                newDimensions.width = this.videoDimensionWidth;
+
+                resizeParams = {
+                    creativeDimensions: newDimensions,
+                    videoDimensions: this.environmentData.videoDimensions,
+                };
+            
+                this.requestResize(resizeParams);
```
Kaleab470, line 72: this whole chunk of code could be shared across the conditionals as well and should be put after the last if statement

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
+const QUARTER_THE_SIZE = .25;
+const THIRTY_FIVE_PERCENT = .35;
+const HALF_THE_SIZE = .5;
+const THREE_QUARTERS_THE_SIZE = .75;
+
+/* This creative expands and collapses when the user hovers over the banner. */
```
Kaleab470, line 6: /**

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+     */
+    sendMessageOnClick_(elementName, message) {
+        const sendMessageFunction = () => {this.simidProtocol.sendMessage(message);}
+        document.getElementById(elementName).addEventListener('click', sendMessageFunction);
+    }
+
+    /**
+     * Adds a hover event listener to the contents of the iframe that expands the iframe.
+     * @param {String} elementName The name of the element.
+     * @param {Event} event The event performed on the element.
```
Kaleab470, line 52: is this an actual type?

examples/creatives/hover_nonlinear.js
kristen-m: I think initializing resizeParams as a let is correct since we're changing the value assigned on line 67

Kaleab470: That is really the only time she uses resizeParams so she can declare it there as a const as per [this comment](https://github.com/InteractiveAdvertisingBureau/SIMID/pull/394/files#r465957763)

examples/creatives/hover_nonlinear.js
juliareichel: fixed

examples/creatives/hover_nonlinear.js
juliareichel: good idea

examples/creatives/hover_nonlinear.js
juliareichel: yes!

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,143 @@
...
+                    resizeParams = {
+                        creativeDimensions: newDimensions,
+                        videoDimensions: this.environmentData.videoDimensions,
+                    };
+                
+                    this.requestResize(resizeParams);
+            }
+
+            //if ad is between the top half and three quarters of the player
+            else if (this.initialDimensions_.y > (this.videoDimensionHeight * HALF_THE_SIZE) &&
```
ryanthompson591, line 92: It feels a little more complex than needed.  I'll have to see how this works in practice.

In general for this type of problem, lookin at where the nonlinear is, might be a wrong path, instead ideally just focus on how much you want it to expand.

That said, I don't think there's a bunch of time for fixing, since the internship is coming to a close.  If you do have more time to make this simpler though, have a try at it.

juliareichel: I drastically improved the logic, take a look at new commits

examples/creatives/hover_nonlinear.js
juliareichel: totally restructured code

examples/creatives/hover_nonlinear.js
juliareichel: outdated

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,141 @@
...
+                return;
+            }
+
+            //Makes sure height fits in the player
+            fullCreativeHeight = desiredY + desiredHeight;
+            if (fullCreativeHeight > this.videoDimensionHeight_) {
+                fullCreativeHeight = this.videoDimensionHeight_;
+                desiredHeight = fullCreativeHeight - desiredY;
+            }
+
```
ryanthompson591, line 107: Yeah, this looks a lot like what we discussed. It should "just work" in all situations. You might not get as good of an effect with the postage stamp non linear, but that could be an edge case.

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,141 @@
...
+            let fullCreativeHeight;
+
+            //Width grows 20% 
+            desiredWidth = this.initialDimensions_.width * DIMENSION_GROWTH;
+            desiredX = parseInt(this.initialDimensions_.x) - parseInt(this.initialDimensions_.width * OFFSET_SHIFT);
+
+            //Makes sure x offset fits in the player
+            if (desiredX <= 0) {
+                desiredX = 0;
+                desiredWidth = this.initialDimensions_.width * DGE_DIMENSION_GROWTH;
```
kristen-m, line 75: *EDGE_DIMENSION_GROWTH

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,141 @@
+const DIMENSION_GROWTH = 1.4;
+const OFFSET_SHIFT = .2;
+const EDGE_DIMENSION_GROWTH = 1.2;
```
kristen-m, line 3: What's the difference between dimension growth & edge dimension growth?

examples/creatives/hover_nonlinear.js
```diff
@@ -0,0 +1,141 @@
...
+            }
+            
+            //Height grows 20%
+            desiredHeight = this.initialDimensions_.height * DIMENSION_GROWTH;
+            desiredY = parseInt(this.initialDimensions_.y) - parseInt(this.initialDimensions_.height * OFFSET_SHIFT);
+
+            //Makes sure y offset fits in the player
+            if (desiredY <= 0) {
+                desiredY = 0;
+                desiredHeight = this.initialDimensions_.height * DGE_DIMENSION_GROWTH;
```
kristen-m, line 95: fix same spelling error here

examples/creatives/hover_nonlinear.js
juliareichel: good catch

examples/creatives/hover_nonlinear.js
juliareichel: EDGE_DIMENSION_GROWTH is applied when the iframe is at the edge of the player (ie: when the x/y offset is 0, or when you try and expand the iframe the calculations would cause it to expand outside the player dimensions). When this occurs we have to make sure the %s are pretty much equivalent to those that occur when the iframe is lets say at the center of the player div and there is room to expand without going over the player's borders (in which case DIMENSION_GROWTH would be applied since the OFFSET_SHIFT is capable of being applied.

examples/creatives/hover_nonlinear.js
juliareichel: fixed

examples/creatives/hover_nonlinear.js
juliareichel: the "postage stamp" expands to just a bigger postage stamp, which I think is good for the purpose of our sample creative

PR #397, opened 2020-08-04
Integrate game with banner ad

Kaleab470:


examples/creatives/banner_nonlinear.js
```diff
@@ -20,8 +20,8 @@ class BannerNonLinear extends BaseSimidCreative {
      */
     addButtonClickActions_() {
         this.sendMessageOnButtonClick_("close_ad", CreativeMessage.REQUEST_STOP);
-        this.sendMessageOnButtonClick_("ad_text", CreativeMessage.REQUEST_EXPAND);
-        this.sendMessageOnButtonClick_("minimize_ad", CreativeMessage.REQUEST_COLLAPSE);
+        this.onExpand_();
```
juliareichel, line 23: if there's a function already meant to send message on-click, why not just utilize that and put your css portions of the code in a callback function instead of creating these 2 new functions?

Kaleab470: I'm not sure what you mean by that- how would that look?

juliareichel: using the sendMessageOnButtonClick_ function and then just perform your classAdds in => function

Kaleab470: that's a great idea! Thanks

examples/creatives/banner_nonlinear.js
```diff
@@ -3,36 +3,101 @@ const X_OFFSET_PERCENTAGE = .15;
...
         this.dynamicResize_();
     }
 
+    /**
+     * Checks validity of ad parameters and rejects with proper message if invalid.
+     * @param eventData an object that contains information details for a particular event
+     *   such as event type, unique Ids, creativeData and environmentData.
+     * @private 
+     */ 
+    validateAndParseAdParams_(eventData) {
```
kristen-m, line 45: Since we're validating the adParams here as well, would we want to add some checks to make sure that (at the very least) the webUrl is included in the adParams? Something like we did with the map creative when we checked if the query was included, maybe making thne url a required param? something like [this](https://github.com/InteractiveAdvertisingBureau/SIMID/blob/master/examples/creatives/simid_map_creative.js#L75)

Kaleab470: yeah that sounds good

examples/creatives/banner_nonlinear.js
```diff
@@ -3,36 +3,114 @@ const X_OFFSET_PERCENTAGE = .15;
...
         this.addButtonClickActions_();
     }
 
     /** @override */
     onInit(eventData) {
-        super.onInit(eventData);
+        this.updateInternalOnInit(eventData);
+        this.validateAndParseAdParams_(eventData);
+        this.updateCreativeWithParams_();
         this.dynamicResize_();
```
damargulis, line 36: If init gets rejected, you probably don't want to continue on to call updateCreativeWithParams and dynamicResize

examples/creatives/banner_nonlinear.js
```diff
@@ -3,36 +3,114 @@ const X_OFFSET_PERCENTAGE = .15;
...
     }
 
+    /**
+     * Checks validity of ad parameters and rejects with proper message if invalid.
+     * @param eventData an object that contains information details for a particular event
```
damargulis, line 41: give eventdata a type

examples/creatives/banner_nonlinear.js
```diff
@@ -3,36 +3,114 @@ const X_OFFSET_PERCENTAGE = .15;
...
+     *   such as event type, unique Ids, creativeData and environmentData.
+     * @private 
+     */ 
+    validateAndParseAdParams_(eventData) {
+        if (this.creativeData.adParameters == "") {
```
damargulis, line 46: nit.
This can also be written as `if (!this.creativeData.adParamaters)` since the empty string evaluates to false in js

examples/creatives/banner_nonlinear.js
Kaleab470: looking at the other creatives and how they were wrote, they all finish executing that function even if it gets rejected. Is there a way to do what you said without adding too much complexity to the creatives?

damargulis: hmm, if the others do it then I guess its fine to do this just to keep it consistent.  I guess if init is rejected the iframe will just never show so it doesn't really matter what else you do

